PR #1, opened 2015-02-07
revive lists site-building code

jbclements:
It appears I have to bring Julius Caesar back to life before killing him--this pull request adds code to build a lists.racket-lang.org site, mostly by reviving code that had been deleted last year. I followed the structure of the "downloads" site, mostly.  I also added some light gumby-ization to make it fit our existing js, and amended links so that the various 'dev' references now point to google rather than to our own server. 

Please note that the new code does not build any of the lists.racket-lang.org subdirectories, including
- listinfo
- dev
- dev/archive
- users
- users/archive
- ...et cetera.

I'm assuming we can just leave the ones that are currently there in place. Let me know if that's not the case.

Also, let me know if you want me to squash. I'm not a big squash-fan, myself, but I'm happy to do it if it's what others want.


rmculpepper:
Merged, thanks. Server updated.

PR #2, opened 2015-05-01
webpage for fifth RacketCon

mbutterick:


samth:
Can you post what it looks like?

mbutterick:
[What it looks like](http://mbutterick.github.io/racket-lang-org/rcon/2015-mb/)

samth:
Awesome!

takikawa:
This is awesome, thanks for doing it! Just an idea: could the picture show the code used to create it when hovered/clicked/something?

rfindler:
Very nice.

Robby

mbutterick:
I like that idea. I'll figure out how to do it then add it later.

stamourv:
It's live, and looking great! Thanks again!
There is a noticeable loading delay for the svg file, though. I see you have also included a zipped version, but I did not get it to work on my browser. I don't know anything about web dev, so I likely did something wrong. Any ideas?

mflatt:
@stamourv: Probably I should change the web-upload mode of `s3-sync` to gzip ".svg" files in the same way that it gzips ".css" and ".html". Then, accessing as the ".svg" name gets the gzipped version --- on the assumption that only modern web clients will ask for the SVG file, and those clients always enable gzip encodings.

mflatt:
The `s3-sync` package is now updated

stamourv:
Ok, I've resynced. It's a bit better, but it's still noticeable. It may be better to just ship a png instead.

mflatt:
FWIW, it now works well for me. Adding an explicit size in the HTML reference to the SVG image might make it slightly better.

mflatt:
Also: +1 on the design

PR #3, opened 2015-05-02
change .svg to .svgz

mbutterick:
The slow-loading SVG was my error. I made an .svgz version and forgot to update the reference inside the page.


mbutterick:
Never mind this, I have a better idea.

PR #4, opened 2015-05-02
simplify / shrink svg image

mbutterick:
I regenerated the SVG using `(line-samples 100)`, which simplifies the curves and thus shrinks the file by 5x. That, combined with the server-side gzipping that @mflatt set up, should make it sufficiently speedy for the time being.


stamourv:
Merged, thanks!

PR #8, opened 2015-07-02
implement popout descriptions

mbutterick:
Descriptions appear / disappear when speaker name is clicked.


stamourv:
Merged, thanks!

mbutterick:
BTW you’ll want to remove any hard linewrapping from the description / bios. They will be decoded so that double linebreaks become paragraphs (<p> tags) and single linebreaks become <br> tags.

stamourv:
Yep, I saw that. I missed a few, though. Fixed locally, and will push as soon as I can. Thanks!

PR #17, opened 2016-06-06
updated design for RC website

mbutterick:
@stamourv: I plan to refine this, but this will get you started.


mbutterick:
(PS the two-tone lettering is generated by a [Racket program](https://github.com/mbutterick/racket-lang-org/blob/227fd60717be10a6d815f8f2651db5529dfa021d/rcon/2016/racketcon-font.rkt), of course)

stamourv:
This is really cool!
The hatching on the sides makes me a bit dizzy, though. Maybe spacing the lines out would help?

florence:
I have a *really* hard time reading the title lettering---whatever pattern matching my brain uses doesn't recognize a lot of the letters.

rfindler:
I like it. Here are two concrete ideas for making it slightly easier to parse, visually speaking:

- make the parens round, or drop them, and maybe drop the word "sixth" at the top

- put a space in "also speaking"

mbutterick:
I just noticed that the color animation doesn't work without HTTP. So [here it is](http://mbutterick.github.io/racket-lang-org/rcon/2016/).

mbutterick:
I lightened the background image.

I could start the lettering in a more legible state, and then let the animations mess it up, rather than the more op-art starting condition.

As for the specific letters / words, that can be whatever you want. No parens on the top — is fine. "ALSOSPEAKING" could just be "SPEAKERS" (the goal is merely to use more letters than "KEYNOTE", thereby forcing the headline to be smaller)

rfindler:
My comment about the parens is that I couldn't figure out what the words were on the top because I didn't recognize the parens as parens. I was trying to figure out what the 9th computer science thing was for a while before I gave up and went onto the next line and only then realized what was going on.

Overall, I like the page a lot and meant the suggestions to make it easier for people to figure out what that first line was.

mbutterick:
Oh I don’t mind the suggestions at all. I just pushed a fix to [give the letters black outlines and simpler shapes](http://mbutterick.github.io/racket-lang-org/rcon/2016/). Better?

rfindler:
Beautiful!

rfindler:
My 9-year old son just wandered by and saw the front page and said "sixth racket con -- what is that? Is that a game? I hope so."  :)

stamourv:
I love it!

Re the "speakers" section: I've been using "confirmed speakers" to convey that this is not the complete list, only the list so far. That's probably too long a title, especially with the large, animated text. Maybe we could just add "and more to come!" at the end?

stamourv:
In any case, I can merge this version if you're happy with it, and do that change myself.

mbutterick:
I just made that change. Yes, if you want to merge this version, I approve.

mbutterick:
PS @stamourv I see you made some changes to the RCon home page since I forked it. You’ll want to bring those across.

stamourv:
New design is merged, and mostly live.
For some reason, the live version insists on using last year's CSS, but a local version works just fine. Investigating.

stamourv:
CSS issue fixed now. Had to purge a cache somewhere.

PR #18, opened 2016-06-08
RacketCon 2016 website tweak

mbutterick:
Sorry, @stamourv — I wasn’t sure where the new site lives in this repo (the `rcon/2016` subdirectory still seems to contain the stub site). But the only commit that matters is the last one: I converted the background to SVG so it looks better on all devices.


stamourv:
I did forget to delete the leftovers from the stub site. Thanks for cleaning them up!

rfindler:
FWIW, when I look at con.racket-lang.org on my iphone, the whitespace on the left is larger than that on the right (horizontal or vertical orientation).

mbutterick:
> FWIW, when I look at con.racket-lang.org on my iphone, the whitespace on the left is larger than that on the right (horizontal or vertical orientation)
> 

That is deliberate for the desktop view, though I agree that the mobile version should have no left margin. It is on my list of tweaks.

PR #19, opened 2016-06-22
RacketCon 2016 website improvements

mbutterick:
* change background stripe to SVG
* add smooth transitions to link hovers
* add style overrides for mobile (margins, columns, etc., as suggested by @rfindler)


stamourv:
Thanks!

PR #22, opened 2016-06-29
convert Racket blog to Frog

mbutterick:


mbutterick:
The `blog.racket-lang.org` should be mapped into the new `blog` directory at the top level of the repo.

I’ve preserved the URL structure from the current blog as much as possible, so existing links to blog articles should still work.

The Markdown sources for the posts is in `blog/_src/posts`. This could be moved elsewhere in the repo if others find it convenient (e.g., to keep the content and infrastructure separate). The rules for adding & editing posts are in the [Frog docs](https://github.com/greghendershott/frog). 

The blog depends on the `frog` and `gregor` packages. Additionally, for syntax highlighting, it relies on Python and `pygments`.

To regenerate the blog, from the command line do `raco frog -b` (that should probably become part of the build script for the website)

samth:
Awesome! We also need to change the `sync.rkt` code to upload it to the correct S3 bucket.

mbutterick:
I know nothing about the S3 integration. I will leave that to wiser minds.

samth:
The s3 integration is pretty simple: we need to add a line to `sync.rkt` around [here](https://github.com/racket/racket-lang-org/blob/master/sync.rkt#L65) which specifies which directory things get built in, and which web site they go to (`blog.racket-lang.org`).

We also need to arrange for the build process (currently controlled by `all.rkt`) to build the blog. Since `all.rkt` uses the `plt-web` library fundamentally, there should probably be a new build process that runs that, plus runs `frog` for the blog. Is there an easy API to use for `frog`?

@mflatt and @stamourv have been the other heavy users of the web site recently -- is there anything else you two are concerned about?

stamourv:
Can't think of anything.

mbutterick:
I’d like to finish up this PR. I need to make a few adjustments to the CSS & update the blog source with the most recent posts. But are we otherwise ready to “upload it to the correct S3 bucket”?

mbutterick:
Superseded by #26.

PR #23, opened 2016-08-12
fix year & day in heading

bennn:
Makes heading agree with the text below it.


stamourv:
Merged. Thanks for pointing this out!

PR #24, opened 2016-08-12
try fixing con2015 resources links

bennn:
@mbutterick 

The RacketCon 2015 links to slides are broken because they all point to `2015/2015/...` e.g. http://con.racket-lang.org/2015/2015/burns.pdf

Here's maybe the first step in fixing.


stamourv:
Merged, along with the other step. Thanks!

PR #25, opened 2016-11-08
tweak subscription text

jbclements:
Make some small changes to the subscription text to reduce confusion about the subscription process.
1) remove the explicit mention of typing an e-mail address--it doesn't make sense for the google groups ones.
2) tidy the google groups text so that it says "Join the <whatever> mailing list".


stamourv:
Merged, thanks!

PR #26, opened 2016-12-10
convert blog to frog

mbutterick:
Supersedes #22.

The `blog.racket-lang.org` should be mapped into the new `blog` directory at the top level of the repo.

I’ve preserved the URL structure from the current blog as much as possible, so existing links to blog articles should still work.

The Markdown sources for the posts is in `blog/_src/posts`. This could be moved elsewhere in the repo if others find it convenient (e.g., to keep the content and infrastructure separate). The rules for adding & editing posts are in the [Frog docs](https://github.com/greghendershott/frog). 

The blog depends on the `frog` and `gregor` and `pollen` packages. Additionally, for syntax highlighting, it relies on `python` and `pygments`.

To regenerate the blog, from the `blog` directory do:

```
raco frog -c   # cleans out previous build
raco frog -b   # builds all HTML files from Markdown sources
pollen render  # updates CSS file from Pollen source
```

I’ve looked at the design in Mac, Windows, Ubuntu, and iPhone (the design now collapses to a single column for mobile things). LGTM but there may still be nits lurking and I welcome bug reports.

`blog` directory layout
-

There’s sort of a delicate mix of source files and generated files here. The subdirectories `_src`, `css`, `img`, and `blog/index.ptree`, are all source-like material that are used to rebuild the blog. There is also a hidden `.frogrc` and `.frog` directory that are used during a build. The other directories & HTML files are regenerated by `frog` when `raco frog -b` runs. 

AFAICT `frog` is somewhat particular about the arrangement of files (for that matter so is `pollen`). From a dev perspective, the advantage of keeping everything in one directory is that it cooperates properly with the `frog` and `pollen` preview tools. If other things are equal, I’d recommend a build process like this:

1. Build blog using script above.
2. Copy the directory.
3. From the copy, delete `_src`, `.frogrc`, and `.frog`.
4. Move this pruned copy onto the live server.

`raw-data`
-

The `raw-data` subdirectory is an XML dump of the current blog, with some Markdown conversion tools. It’s strictly archival: it is not used in generating the blog. (Nor can it be used to regenerate the Markdown sources, as I went through them and hand-edited them in detail.)


mbutterick:
Also, this redesign was already cleared for takeoff some months ago. So even though I’m working on updates for the other web pages, I see no reason to hold this back (In fact for me it would be preferable if things were pushed live in stages so that the waves of public complaint are staggered). Thanks dudes

samth:
I think we should do this ASAP. However, I'd prefer not to have generated files checked in, which this has. 

We will want have a build script to copy this to an s3 bucket named `blog.racket-lang.org`, using a script like the one here: https://github.com/racket/racket-lang-org/blob/master/sync.rkt#L49-L66

mbutterick:
>  I'd prefer not to have generated files checked in, which this has.

OK, I can take these out.

> We will want have a build script

OK, I’ll see how far I can get with this.

mbutterick:
BTW, why does this need to go through S3 at all? Wouldn’t it be simpler to move it to a separate repo and host it through GH pages? Now that it’s on Frog, it’s just static pages.

samth:
We host all of our other static sites on S3, so it seems simplest to continue that.

mbutterick:
Ah, well you did ask for “ASAP”. I suggested GH pages because it’s a lot closer to ASAP than the S3 route. But as you wish.

mbutterick:
When I try to run [`sync.rkt`](https://github.com/racket/racket-lang-org/blob/master/sync.rkt) even in `dry-run?` mode, I get an error that I’m missing the file `/Users/MB/.aws-keys`. I assume this is some kind of permissions file that authorized users have. So I will not be able to test that part.

stamourv:
When the repo is in a state where it's ready to be synced, I can do it for you.

mbutterick:
Each of the existing S3 uploads has a hostname target associated with it, e.g., `(upload "www" "www.racket-lang.org")`. Is there a `blog.racket-lang-org` on S3 ready to receive the blog files?

samth:
Yes, I created that bucket a while back.

mbutterick:
OK, I have adapted the `rcon/all.rkt` script into a new `blog/all.rkt` that registers a blog site called "blog" and seems to copy all the files correctly. (I’m afraid `plt-web` will only be getting a two-star Yelp review.)

What is the appropriate way to trigger the `frog` and `pollen` parts of the build? From within `all.rkt`? I guess that’s possible, but they’re most readably handled with `raco` commands. 

I was hoping for guidance from `rcon` but found this message instead:

```
;; TODO build the pollen file (and the SVGs) when the site is built, and remove
;;  the generated files from the repo
```

mbutterick:
I suppose I could do this within `blog/all.rkt`, but how would I get the full `raco` path?

```
(system "raco frog -b")
(system (format "raco pollen render ~a" (build-path blog-dir "css")))
```

rfindler:
It might be safe to use `find-exe` and the expect `raco` to be in the same
directory as `racket` (the result of that command). There may be a better
way, tho.

blog/all.rkt
```diff
@@ -9,6 +10,13 @@
         #:url (rewrite-for-testing "https://blog.racket-lang.org/")))
 (register-identity blog-site)
 
+(define-runtime-path blog-dir ".")
+(define css-dir (simplify-path (build-path blog-dir "css/")))
+(define raco-path (let-values ([(base name _) (split-path (find-exe))])
+                    (build-path base "raco")))
+(system (format "~a frog -b" raco-path)) ; frog rebuild: generates blog html
+(system (format "~a pollen render ~a" raco-path css-dir)) ; pollen rebuild: generates css
+
```
rfindler: It is probably wise to use `system*` here. And if this might run on a windows machine, use `raco.exe` (but maybe that doesn't matter).

mbutterick: Swapping `system` for `system*` doesn’t work for me. 

```
racket: exec failed (No such file or directory; errno=2)
```

samth:
I think this transcript shows the right way to run `raco` commands:

```
[samth@hermes:~/sw/plt (master) plt] r
Welcome to Racket v6.7.0.3.
> (require raco/all-tools)
> (define v (all-tools))
> (current-command-line-arguments (vector "-V"))
> (dynamic-require (second (hash-ref v "frog")) #f)
Frog 0.25
Configuration /home/samth/sw/plt/.frogrc not found; using defaults.
Very verbose mode
> 
```

rfindler:
Yes, @samth 's way is better, I agree!

As for `system*`, you would have to pass the path to the binary and the path to the arguments separately.

`system*` talks directly to the OS to run the program instead of going thru a shell (like bash) and thus is not vulnerable to strange things like interpreting random environment variables (like a broken/malicious path environment variable, etc).

samth:
Just since I've been doing this, here's how to run those two commands (not tested):

```
(require raco/all-tools)
(define v (all-tools))
(parameterize ([current-command-line-arguments (vector "-b")])
  (dynamic-require (second (hash-ref v "frog")) #f))
(parameterize ([current-command-line-arguments (vector "render" css-dir)])
  (dynamic-require (second (hash-ref v "pollen")) #f))
```

mbutterick:
Thanks for the help.

So this PR is ready to go, at least as far as I can test it. 

What I did was start with a clean checkout of this branch and then did this command, which seemed to be the best way to verify the `plt-web` render & copy operations:

```
racket blog/all.rkt -rf -o ~/Desktop/tmp/
```

The `frog` and `raco` rendering happened, all the right files got copied, and all the source files got left behind. Then I tried the top-level `all.rkt`:

```
racket all.rkt -rf -o ~/Desktop/tmp/
```

This time, all the other sites were rendered too, plus the blog.

samth:
Unfortunately, when I did this a while back, I created the bucket in the wrong region, so it had to be deleted, and it appears that I can't create it again (with the right name) for a little while.

Here it is at a temporary URL: http://blog-test.racket-lang.org.s3-website-us-east-1.amazonaws.com/

It looks excellent!

One more thing before we publish this -- do we want to write a blog post about the new blog look (perhaps mentioning the excellent people and tools involved)?

mbutterick:
I’ll write it up.

rfindler:
Our blog is pretty boring these days, so I think it would be great to put
something there.

samth:
Maybe we can copy-paste @mflatt's commit messages about `for` loops to make a blog post. ;)

rfindler:
Oh, yeah! I think that explaining the space safety in `for` loops and in
`map` in a blog post would be great.

samth:
Bucket now correct, and available at http://blog.racket-lang.org.s3-website-us-east-1.amazonaws.com/ .

Remaining tasks:

[ ] Write announcement blog post (@mbutterick)
[ ] Re-direct DNS (@samth)

Anything else missing?

mbutterick:
> Write announcement blog post (@mbutterick)

Just pushed

samth:
@rfindler @stamourv Since you're the other people who've looked at this, anything else you want to see here?

rfindler:
Nope. LGTM.

samth:
The blog should now be live at https://blog.racket-lang.org/

Thanks, @mbutterick, it looks great.

samth:
For completeness, I'll note here one more thing I did: I manually added using the AWS console a redirect from the old Atom feed URL (http://blog.racket-lang.org/feeds/posts/default) to the new Atom feed URL (http://blog.racket-lang.org/feeds/all.atom.xml) and tested that it doesn't go away on sync.

mbutterick:
Maybe it can be fixed. I paid attention to preserving the permalinks for the blog posts themselves, but I concede I didn’t look at the RSS URLs. What were they before?

samth:
Inconveniently, the RSS url is http://blog.racket-lang.org/feeds/posts/default?alt=rss (Docs [here](https://support.google.com/blogger/answer/97933?hl=en))

Perhaps the S3 rewrite rules can be made to do it, but I don't see how: http://docs.aws.amazon.com/AmazonS3/latest/dev/HowDoIWebsiteConfiguration.html

mbutterick:
I know nothing about S3 but apparently a setting called `CloudFront > Behaviors > Forward Query Strings` [must be enabled](http://stackoverflow.com/questions/30775068/amazon-s3-redirect-rule-preserve-query-params) for the query strings to reach S3 (and therefore the redirect rule), which I assume would look like so:

```
<RoutingRules>
  <RoutingRule>
    <Condition>
      <KeyPrefixEquals>/feeds/posts/default?alt=rss</KeyPrefixEquals>
    </Condition>
    <Redirect>
      <ReplaceKeyPrefixWith>/feeds/all.rss.xml</ReplaceKeyPrefixWith>
    </Redirect>
  </RoutingRule>

  <RoutingRule>
    <Condition>
      <KeyPrefixEquals>/feeds/posts/default</KeyPrefixEquals>
    </Condition>
    <Redirect>
      <ReplaceKeyPrefixWith>/feeds/all.atom.xml</ReplaceKeyPrefixWith>
    </Redirect>
  </RoutingRule>
</RoutingRules>
```

I suppose this could also be fixed by putting a dummy file to respond to requests, but the server-side redirect is tidier I think.

samth:
We don't use Cloudfront, so I don't think that's the problem, but I added those rules and they don't fix it for RSS.

mbutterick:
What about adding an error code to the condition? It [has been suggested](http://stackoverflow.com/a/24001725/1486915) that this will preserve the query string.

```
<RoutingRules>
  <RoutingRule>
    <Condition>
      <KeyPrefixEquals>/feeds/posts/default?alt=rss</KeyPrefixEquals>
      <HttpErrorCodeReturnedEquals>404</HttpErrorCodeReturnedEquals>
    </Condition>
    <Redirect>
      <ReplaceKeyPrefixWith>/feeds/all.rss.xml</ReplaceKeyPrefixWith>
    </Redirect>
  </RoutingRule>

  <RoutingRule>
    <Condition>
      <KeyPrefixEquals>/feeds/posts/default</KeyPrefixEquals>
      <HttpErrorCodeReturnedEquals>404</HttpErrorCodeReturnedEquals>
    </Condition>
    <Redirect>
      <ReplaceKeyPrefixWith>/feeds/all.atom.xml</ReplaceKeyPrefixWith>
    </Redirect>
  </RoutingRule>
</RoutingRules>
```

samth:
I tried that (it's currently in the specified state) and it didn't work -- gives a 403 for both pages after I moved the actual `/feeds/posts/default` out of the way.

mbutterick:
Hmm, what about changing the code test to

```
<HttpErrorCodeReturnedEquals>403</HttpErrorCodeReturnedEquals>
```

mbutterick:
(that is, catch a `403` error not a `404`)

samth:
Somehow, that makes the `?alt=rss` version fetch the `all.atom.xml`, and the plain version give a 403. Sigh.

mbutterick:
I suggest we wait until an actual human being asks “what happened to the RSS feed” before continuing.

samth:
I concur. I'll put the atom feed back tomorrow.

samth:
And done.

Issue #27, opened 2016-12-12
delete the logo bar on docs.racket-lang.org


mbutterick:
Rather than change this to match the new logo, I would prefer to delete it (though I’m not sure where in this repo it’s controlled, or if it's elsewhere.)

No reasonable human will be confused by the loss of this ’90s wayfinding artifact. See, e.g., [the Rust docs](https://doc.rust-lang.org/book/getting-started.html).

stamourv:
I agree.
One thing I do like about the bar, though, (which probably doesn't apply to the Rust docs) is that it makes it immediately obvious that one is looking at the online version of the docs, rather than a local one. Probably not indispensable, though.

mfelleisen:
I will miss it and the shield. :((((

mbutterick:
I would be open to the idea of amending the CSS for the docs themselves to behave differently when they're displayed on racket-lang.org (e.g., put a small logo & link to the main home page), so there’s a visual distinction from local builds.

But the current header strip would still need to be clipped off (it is being added to the docs from outside)

stamourv:
Sounds like a good solution to me.

mbutterick:
How is the toolbar prepended to the docs pages? Does it have something to do with [this file?](https://github.com/racket/racket-lang-org/blob/master/stubs/docs.rkt) Obviously I don’t want to just take a machete to things. (Actually, that’s not true. I do. But prudence stays my hand.)

Once that’s removed, I will do a PR for `racket/scribble` that updates the CSS for the docs.

mflatt:
Building the web pages creates "stubs/docs", or something like that, containing "doc-site.css" and "doc-site.js". Those two files get uploaded to docs.racket-lang.org manually.

To test out a change, you can copy those two files to the directory of a local build (to replace the stub files there).

mbutterick:
Aha, so that’s what those two files do. OK, maybe I can just do it all the surgery as one PR here.

mbutterick:
http://docs.racket-lang.org/guide/index.html

I prefer how this looks. Any dissenters?

In Mac OS Chrome I see a little flash on each page when the logo loads (I didn’t see this on my local test) though not in Safari or Firefox, so maybe it’s just a me glitch, not a world glitch.

rfindler:
It looks great on my browser, but strange on my phone. (image attached; ios safari, latest version as far as I know)

[...]

PR #28, opened 2016-12-17
update Racket home page

mbutterick:
Updates the Racket home page to the [version seen here](http://mbutterick.github.io/racket-lang-org/www/).

There is a [related pull request](https://github.com/racket/plt-web/pull/1) for `plt-web`. 

My technical goal was to alter as little of the build process as possible (and thus minimize the risk of breaking things, both in terms of the build, and in terms of preserving existing URLs that point into the website). AFAIK I have only replaced one HTML file (`index.html`) and added one other (`plt.html`). All the rest remain as they were.

Well, almost. To make the existing pages more consistent with the new visuals, I changed the central logo images & CSS in `plt-web`, which in turn get applied to the `plt-web`-generated pages in `racket-lang-org`. I pursued the same minimal strategy: changing as little CSS as possible, and keeping the images the same size. 

This way, we can at least get the new design rolling, and dig into the lower-level pages later on. (Also, it’s not obvious to me which ones are vestigial, and can be ignored, vs which are current.)

I’ve tested this build locally using `racket all.rkt -rf -o ~/Desktop/tmp/` at the top level, and everything seems to work properly, and I get no errors.

PS. It’s not clear to me how this repo or `plt-web` influence http://docs.racket-lang.org/ or http://pkgs.racket-lang.org, so that part I haven’t tested. But if they automatically pick up their logo and toolbar from `plt-web`, then they should be updated automatically.

PPS. This does not supersede quibble #27 — I still don’t think http://docs.racket-lang.org/ should have a logo bar.


samth:
There seem to be some weird interactions with tapping on the top set of boxes on my phone browser (Firefox Nightly) (since you made the boxes bigger). They don't go away when I tap on them, or on something else, except then sometimes they do, and sometimes a different one than the one I tapped goes away.

AlexKnauth:
This looks bad, with overlapping text, on a phone screen

samth:
@AlexKnauth can you be more specific? It works nicely on the mobile devices I've tried.

rfindler:
It works nicely on my iphone's safari, FWIW.

AlexKnauth:
Down by the examples. The description of how "The 2htdp/image library provides easy-to-use functions for making images" overlaps with the next example, with Scribble. The Scribble description overlaps with the next example, and so on

samth:
@AlexKnauth I can't make that happen with any setting I try in either the Firefox or Chrome dev tools for showing various widths, nor on my phone. What phone/browser/screen width do you have, and can you post a screenshot?

mbutterick:
> screenshot

Yes, screenshots for phone-related complaints are the only practical way I can theorize about the cause of these issues, otherwise I am left to imagine the horror

samth:
Also, the issue I had with clicking on the boxes reproduced for me in the Chrome dev tools when I had a narrow-enough width that everything is single column.

AlexKnauth:
This is on my iPhone screen with Safari:
image

rfindler:
@AlexKnauth do you have the latest version of your ios software?

AlexKnauth:
> On Dec 17, 2016, at 6:36 PM, Robby Findler <notifications@github.com> wrote:
> 
> @AlexKnauth do you have the latest version of your ios software?
> 

No, I do not. I guess that means it's been fixed for later versions

lexi-lambda:
Hmm. This definitely looks very pretty, but there are some things about this design I’m not super fond of. Specifically, it feels sort of confusing compared to the existing design, and it doesn’t communicate some things quite as well.

### First comments

Given that the home page is most relevant to people who haven’t used Racket before (and likely don’t have a strong idea what it is), it seems important to almost immediately establish what Racket is and why you should care about it. The old site answers the “what is this?” question absolutely immediately, with the first, largest line of text on the page: **A programmable programming language**. The following blurb is also the *only* longform piece of text above the fold, it doesn’t compete with anything else in that part of the page, and it makes its point extremely quickly. I think the paragraph on the home page is a great condensation of @mfelleisen’s Racket Manifesto.

The new site tells me very little up front, and the above the fold content is a lot more cluttered. The closest thing to the old design’s top heading is the slogan-y “solve problems, make languages”, which is both a lot more vague *and* set in a more subdued color, so it doesn’t really attract the eye. The single paragraph has been split into a set of boxes with headings, but the headings don’t really get across the same information that the paragraph did.

### Info boxes

The boxes themselves are probably the most confusing part of the new design to me, which isn’t good, given that they’re the first thing on the page and occupy most of the above the fold area on my 15” laptop display. The boxes themselves show nothing at all until you hover over them, which is somewhat poor UX in general given that it makes the *primary content* of the site something the user has to expend effort to discover, but it’s *especially* bad on mobile, where mouse hover isn’t a thing to begin with. Tapping on the boxes is sort of a replacement, but it’s possible to actually tap on a part of the box that is a link, which takes me away from the homepage, despite the fact that I couldn’t know I was going to click a link. Given that all the boxes are vertically aligned on mobile, anyway, the hover-to-show behavior should *definitely* not be present on mobile devices.

Even on desktop, hiding the content is confusing. It would be better if there were stronger clues about what was inside each box, so that just glancing at them would give a general overview, but that isn’t the case—the headings are alright, but they’re not terribly informative (what makes Racket macros powerful? why language**s**, plural? is being cross platform really a selling point? what does “the best of Scheme and Lisp” mean?), and the graphics are completely meaningless.

I disabled the hovering functionality to see what the site would look like if all the boxes were visible by default, and the result is still confusing, but in a different way. Now, instead of seeing nothing, I see too much. Instead of a single call to action (which the old design has), I see at least six, plus a deluge of links. Still, I think this is actually considerably better than hiding all the information and asking people to move a cursor around to see it, and it would be improved by a bit more spacing to [make the groupings more clear](https://en.wikipedia.org/wiki/Principles_of_grouping) and to appear less overwhelming.

Clearly, I’m not fond of the boxes. But I’m not against them in principle; I just think they’re not a very good thing to put *first*. A newcomer to Racket should ideally have 1-3 calls to action, not 24, and the call to action should be easily visible and not hidden behind some interactive element. I’d probably be less opposed to the boxes if they were (1) pushed down on the page, below the code samples, and (2) made visible by default instead of on hover.

### Code snippets

These are a lot better. In fact, I want to start by talking about some of the things I really like! I like that the first example is one that draws a picture; I think that’s a much more interesting one than a much more verbose way to write `find . -name *.rkt`. I also really like that the examples are all visible at once instead of having to click through a carousel, which makes it a lot easier to just skim through them and see what Racket has to offer. Highlighting the different `#lang`s is really smart, too, since it drives home why `#lang` is so cool.

However (of course), there are things I dislike, too. Once again, the examples are pretty much impossible to comfortably read unless you move your mouse over them, which defeats a lot of the purpose of having them all visible at once. Also, the examples have lost their code highlighting, which I guess is to fit the new, monochrome color scheme better, but it makes them a lot harder to visually scan, especially for someone who might not be familiar with Lisp syntax.

The examples are also just a lot physically smaller (that is, wrt font size) than the example on the old Racket home page, and again, there are six instead of just one, which sort of dilutes what people see. My suggestion would be to rearrange these elements: put the `#lang racket` example first, make it larger, and center it, similar to the way the existing example looks on the existing home page. Maybe even include a picture of what the program’s output is to make it clear what the program actually does. Then, include the other language examples below.

Remove the on-hover change on mobile, and on desktop either remove it as well or make it a *lot* more subtle. Consider adding a heading to each section to give a short explanation about what each example represents: for `#lang racket`, “General-purpose programming”, for `#lang typed/racket`, “Static typing”, for `#lang scribble/base`, “Typesetting prose and documentation”, etc. Either align them all in a single column, or maybe add a seventh snippet to keep the grid: maybe `#lang at-exp racket` to give a small example of language composition?

### Links

The final section of this design is a *lot* of links. This isn’t so bad, but I think the old design still wins in usability here (if not in visual appeal). The “For getting started” and “For experienced Racketeers” headings are super useful for understanding a *progression* of tools; the new design is much more confusing. If I’m new to Racket, which link do I click? I think keeping the “Software” and “Documentation” blocks as separate sections from the link dump would make it possible to give them more useful unique designs that express things in a more tailored way (which the existing site already does).

### General things

This new design is very careful to stick to a monochromatic color palette, using one shade of blue plus a hover shade, a black, and a gray. The old site wasn’t nearly as afraid to venture into a much larger spectrum of colors, which was admittedly often to its detriment. That red used by some of the headings is *searing*. That said, having two primary colors instead of one *does* help with visual grepping, since it cooperates better with those [aforementioned principles of grouping](https://en.wikipedia.org/wiki/Principles_of_grouping). The subheadings are red, and the main headings are big and black. This also makes the old site pop a lot more than the new one, and while that’s not always a great thing (sometimes it’s rather overwhelming), the new site feels so clean that it ends up feeling devoid of personality.

I don’t want to whine here, of course—you’re doing professional design work for free, and a lot of it is so *obviously* better than what currently exists. The [new blog design](https://blog.racket-lang.org) is *gorgeous*, and the minimalist design and color scheme works great for something like that. Having a consistent visual style is important, and the Racket home page could certainly use some love, but I think this design misses the mark in a few places. That’s not to say I think it’s all bad—quite the opposite, I think it’s a wonderful start—but I think it needs some tweaks to make it as functional as the existing site.

As a final note, I want to mention the [Guile Scheme website](https://www.gnu.org/software/guile/), which recently got overhauled, and I think it does a lot of things right. I don’t think it’s perfect, by any means, and Guile is a very different project from Racket, but the homepage nails the “what is this and why should I care?” aspect immediately.

The Guile site uses the oh-so-trendy “brochure” style of design, and honestly, I think it misses the mark in a few places because of that, so I’m not advocating for that style of design. What I *do* want to point out, though, is two things: (1) how clear the mission statement is to anyone who visits the page and (2) how **linear** the site is. As a newcomer to Guile, I can start reading at the top, and I can just scroll towards the bottom and learn as I go. As a newcomer, I don’t want to have to make decisions: those are scary and overwhelming to me! Yes, obviously links are okay and appreciated if I want to explore, but if nothing specific catches my eye, I’d like the site to tell me, without ambiguity, where I should start reading and where I should go from there. I think the old Racket home page has that, but it’s currently missing from this design.

Apologies in advance if any of this came off as nitpicky or unappreciative. You’ve done an immeasurable amount for the Racket community, and I’ve mostly been silent when I ought to have been showering you with praise and thanks. Take from this what you will, and I certainly won’t dig my heels in if you disagree with some (or most) of the things I’ve said.

Thanks for all the hard work,
Alexis

mbutterick:
> I don’t want to whine here, of course

Of course. But broader “spiritual“ critiques are probably best made privately to Racket central command. Because those decisions are ultimately theirs to make.

This thread is available for concrete technical problems.

samth:
>  But broader “spiritual“ critiques are probably best made privately to Racket central command. Because those decisions are ultimately theirs to make.

Without wanting to speak on behalf of anyone else, I strongly disagree with this. The design here (which I'm very excited about) is your work, and the design decisions are yours to make. You should feel free to take on-board any, all, or none of @lexi-lambda's comments, or mine, or anyone else's. But it's your design, and you get to make the choices.

mfelleisen:
Can we postpone the decision till after the G20 summit in Jan?

mbutterick:
> the design decisions are yours to make

If that’s the case, then let’s ship it (net of necessary technical corrections) & let the public respond. People will compliment some things. They’ll complain about others. Great! That’s what will make the design better, faster: testing it. I cover this in my upcoming book, *How to Program Designs*.

lexi-lambda:
With all due respect, while I would enthusiastically agree for almost any other page of the site, I think the homepage is not something to be experimented with so lightly, especially since the people for whom it matters most will be the least suited to provide feedback. People likely to compliment or complain are those who are already bought into Racket—they’re on the mailing lists, they know about this repo, etc.—and those are the kind of people who don’t need the homepage.

The sort of people who need to understand what the heck this funny thing called Racket is aren’t going to have any obvious option to *provide* feedback, even if they have feedback to give. Even if there *was* some obvious “click here to provide feedback about this page” link, they’re the least invested group of people, so they’re highly unlikely to want to expend the effort. This is an area where I think, if you want to do the “let’s just see what people think” approach, you’d need to actively seek out feedback from that group of people, not passively hope that people provide it.

mbutterick:
> People likely to compliment or complain are those who are already bought into Racket … those are the kind of people who don’t need the homepage

Yes, exactly. This is why I say we should “let the public respond”, and as you say, not make it an inside-baseball discussion among Racketeers.

> The sort of people who need to understand what the heck this funny thing called Racket is aren’t going to have any obvious option to provide feedback

We have traffic logs, download logs, etc. that will show whether the redesign positively affects behavior. This is part of “let the public respond”.

And internet outrage always finds an outlet.
 
https://news.ycombinator.com/item?id=6094946
https://news.ycombinator.com/item?id=8206038
https://news.ycombinator.com/item?id=9268904
https://news.ycombinator.com/item?id=9063177
https://news.ycombinator.com/item?id=8372015
https://news.ycombinator.com/item?id=10029347
https://news.ycombinator.com/item?id=9229932

mbutterick:
Committed, but will not go live until someone with AWS credentials (⊉ me) runs the build & sync script.

mflatt:
Uploaded

PR #30, opened 2016-12-20
Missing dependencies

mbutterick:
Working from a `base` installation, I needed to manually install `lazy` and `datalog` to build the website, because of the code samples on the home page.

### DIFF
--- a/info.rkt
+++ b/info.rkt
@@ -20,6 +20,8 @@
                "txexpr"
                "gregor-lib"
                "frog"
+               "lazy" ; for home page sample
+               "datalog" ; for home page sample
                ("s3-sync" #:version "1.9")))
 
 (define pkg-desc "Sources for http://racket-lang.org")


AlexKnauth:
Should those be build-deps, or does this work differently?

samth:
No, I think they're right as regular deps, because you need to have them to run `all.rkt`.

rfindler:
Does running "raco setup" (just like that, no args) not print out the
missing dependencies?

samth:
I think the relevant code ([here](https://github.com/racket/racket-lang-org/blob/a6cf061e8f8c4c6f51a10c52299a7f42052912bd/www/index.rkt#L251)) is effectively the same as `(eval '(require lazy))` and so isn't caught by `raco setup`.

rfindler:
Maybe the docs linking could leave behind the right clues. Probably a job
for another day tho.

mbutterick:
Obsolete.

PR #32, opened 2016-12-24
Fix x-height of monospaced font

leafac:
The monospaced font didn’t match the x-height of the surrounding text.

Before:

before

After:

after

### DIFF
--- a/www/css/styles.css.pp
+++ b/www/css/styles.css.pp
@@ -110,7 +110,7 @@ top-section {
     opacity: .3;
 }
 
-.lang > .inner tt {
+.lang > .inner :first-child tt {
     font-size: 85%;
     white-space: pre;
 }


stamourv:
@mbutterick: Any thoughts?

mbutterick:
I agree with the change, but I’m going to implement it a slightly different way. Thanks.

leafac:
@mbutterick Thank you 👍 And thanks for the new Racket website, it looks great!

mbutterick:
Thanks @leafac.

@stamourv: something went awry in the S3 sync, as the freshly generated `styles.css` file did not get uploaded (but the new `index.html` did).

stamourv:
Looking into it as soon as my Racket is rebuilt.

stamourv:
Hmm. S3 is serving the right version, and I purged the cloudflare cache for that file, but no luck.
@mflatt, @samth: Anything I missed?

PR #34, opened 2016-12-28
Correct link to version 6.7

sondr3:
The link in the news section wrongly links to the blog post for Racket version 6.3 instead of 6.7


sondr3:
Oh, seems like Emacs added a newline to the end of the file without me noticing, let me know if this is something I should change.

mbutterick:
Thanks for the fix.

PR #41, opened 2017-01-09
fix typos in markdown conversion

AlexKnauth:


samth:
I still need to push this to the live site.

PR #43, opened 2017-01-17
Changed link to H2DP from https to http

m0nastic:
The HTTPS link goes to the Pyret language homepage (presumably hosted on the same server), instead of the H2DP homepage

### DIFF
--- a/www/index.html.pm
+++ b/www/index.html.pm
@@ -228,7 +228,7 @@ Community
 Books
 
 
-◊link["https://www.htdp.org/"]{How to Design Programs}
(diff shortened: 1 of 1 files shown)


samth:
We should just fix where the link goes.

m0nastic:
That makes more sense, but I wasn't sure if that was in your control (I wasn't sure who's server that was).

Should I nuke this and open an Issue?

rfindler:
I believe that @shriram is looking into this already.

m0nastic:
Awesome. Thanks.

PR #33, opened 2016-12-28
introduce midpage slug

mbutterick:
One theme in the home-page feedback thus far is that it lacks a pithy “here’s what Racket is” sentence. OK internet, fair enough. I invite you to consider the text & presentation of this PR:

http://mbutterick.github.io/racket-lang-org/www/

Why put it in the middle? Because then it can also introduce the `#lang` boxes, which are unique to Racket (and thus important to show, but for the same reason could arguably be confusing)

Those who dislike the sentence, I invite you to propose better. With one condition: please *please* **please** can we steer clear of opaque jargon and generic claims. As David Ogilvy said, you can’t bore people into buying your product. 

IMO we should have something that would make sense even to the average *Realm of Racket* reader — and maybe even sound like … oh I don’t know … fun?

PR #48, opened 2017-02-07
Publish "On Application" blog post

jackfirth:
WIP, this is a first rough draft. Feedback, comments, concerns, questions, and squirrel feeding tips are all welcome.

This was born out of a discussion with @stamourv, where I suggested using the Racket blog to highlight user packages more often. A few regular articles here and there pointing out new and useful packages could go a long way to building Racket's community, since more and more interesting developments in Racket are occurring outside the main distribution. This post discusses the magic of `#%app`, along with a few packages that overload it for various purposes.

I was tempted to mention my `compose-app` package, but that seems a bit disingenuous. Suggestions for other packages that do things with `#%app` worth mentioning are especially welcome.

### DIFF
--- a/blog/_src/posts/2017-02-05-on-application.md
+++ b/blog/_src/posts/2017-02-05-on-application.md
@@ -5,27 +5,28 @@
 *posted by Jack Firth*
 
 Today I wanted to write about function application. Specifically, how to redefine and customize
-application with Racket. We'll also look at some Racket packages the define interesting and useful
+application with Racket. We'll also look at some Racket packages that define interesting and useful
 forms of function application.
 
 ## The theory
 
-Application defines one half of *lambda calculus*, the formal model underlying much of modern
-functional programming. The other half is *abstraction*, which creates new functions. Creating and
-applying functions lies at the heart of Racket and many other functional languages.
+Application is the process of combining a function and arguments to evaluate a result. Application
+defines one half of *lambda calculus*, the formal model underlying much of modern functional
+programming. The other half is *abstraction*, which is creating new functions. Creating and applying
+functions is the heart of Racket and many other functional languages.
 
 So how are functions applied in Racket? What makes `(if (< 5 10) 'foo 'bar)` a macro use and
 `(< 5 10)` a function use?
 
-That's actually a trick question, because function application *is* a macro in Racket. During macro
-expansion, the Racket expander inspects the first element of an expression to determine if it has a
-binding to a macro. If it doesn't, rather than assume the expression is a function application, the
-expander inserts an *artificial* identifier named `#%app`[racket] into the expression. So in the
-above example, the expression `(< 5 10)` is converted to `(#%app < 5 10)`. This `#%app`[racket]
-identifier doesn't refer to a single specific `#%app`[racket] like the `if`[racket] refers to
-`if`[racket] from `racket/base`[racket], rather it refers to whatever the enclosing environment
-defines `#%app`[racket] to be (which by default means ordinary function application from
-`racket/base`[racket]).
+That's actually a trick question, because function application *is* a macro in Racket. During
+[macro expansion][1], the Racket expander inspects the first element of an expression to determine
+if it has a binding to a macro. If it doesn't, rather than assume the expression is a function
+application, the expander inserts an *artificial* identifier named `#%app`[racket] into the
+expression. So in the above example, the expression `(< 5 10)` is converted to `(#%app < 5 10)`.
+This `#%app`[racket] identifier doesn't refer to a single specific `#%app`[racket] like the
+`if`[racket] refers to `if`[racket] from `racket/base`[racket], rather it refers to whatever the
+enclosing environment defines `#%app`[racket] to be (which by default means ordinary function
+application from `racket/base`[racket]).
 
 However, imported modules can provide their own definitions of function application by providing an
 `#%app`[racket] macro. Let's define our own `#%app`[racket] that, in addition to applying a
@@ -102,7 +103,8 @@ arguments to functions like `map` and `filter`:
 The `rackjure`[racket] package redefines `#%app` to make working with nested dictionaries easier.
 Dictionaries can be used to get and set values for keys when used as procedures, and when
 dictionaries are the *second* value of a function application the first value is interpreted as a
-key and the dictionary's associated value is looked up. See [the documentation][2] for details.
+key and the dictionary's associated value is looked up. See [the `rackjure` documentation][2] for
+details.
 
 [1]: https://github.com/samth/fancy-app
 [2]: http://docs.racket-lang.org/rackjure/index.html#%28part._dict-app%29
--- a/blog/_src/posts/2017-02-05-on-application.md
+++ b/blog/_src/posts/2017-02-05-on-application.md
@@ -84,7 +84,7 @@ Applying #<procedure:+> to arguments '(6 11)
 
 Redefinition of `#%app`[racket] is occasionally used when defining new languages, but a much more
 pedestrian use is to add some notational shorthand to make certain constructs more convenient. For
-example, consider the [`fancy-app`][1] package. This package provides an `#%app`[racket] macro that
+example, consider the [`fancy-app`][2] package. This package provides an `#%app`[racket] macro that
 behaves just like normal function application *unless* one or more underscores are used. In that
 case, the function application is converted to a lambda that takes as input one argument for each
 underscore. For example, `(format "Hello ~a" _)` is equivalent to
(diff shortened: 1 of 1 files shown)


mbutterick:
Is there a name for the class of identifiers prefixed with `#%`, like `#%app` and `#%datum` and `#%module-begin`?

I don’t understand the “`if` [racket]” notation for certain identifiers used inline.

>I was tempted to mention my compose-app package, but that seems a bit disingenuous

?? Of course you should. Always Be (Self) Promoting.

AlexKnauth:
I noticed you mentioned `fancy-app` and `rackjure`, with a possible addition of `compose-app`. Why wouldn't `lazy`, `typed/racket`, other typed languages created with Turnstile, or `rosette` be included?

bennn:
> Is there a name for the class of identifiers prefixed with #%, like #%app and #%datum and #%module-begin?

I think "core forms" is the right term, even though "core forms" includes "non-`#%` things" like `module` and `define-values`
http://docs.racket-lang.org/reference/syntax-model.html#%28part._fully-expanded%29

lexi-lambda:
This might be an unpopular opinion, but I think libraries like `fancy-app`—things that export `#%app` without being `#lang`s—are naughty/hacky, abuse the purpose of `#%app`, and should not be encouraged. Like `#%top`, `#%datum`, and `#%module-begin`, `#%app` is fundamentally a hook for *module languages* to customize the behavior of a Racket core concept. They are not things meant to be tweaked otherwise.

Why? Well, for one, they don’t compose. Only one `#%app` can ever be in effect at a time. There’s a more insidious problem, though, which is that overriding a `#lang`’s `#%app` can break invariants that the language is looking to maintain by using a custom `#%app`. Turnstile-built languages are an example of this, since they do typechecking as part of the expansion of `#%app`. Using a different `#%app` will at best cause type errors due to a failure to propagate type information and will at worst actually violate the type system and cause runtime errors and unsoundness.

One possible solution to this problem would be to make `#%app` a “mixin”, something that always expands to the `#lang`’s `#%app`, and you could layer multiple `#%app`s on top of each other. I think @jeapostrophe may have mentioned doing something like this with `remix`, but I don’t remember. That said, such a thing does not currently exist, so completely replacing `#%app` seems like a bad practice to me.

**tl;dr**: Module languages *should* provide their own `#%app` if they need custom function application behavior. Libraries *should not*.

jeapostrophe:
The thing I mentioned was a require* that is like let* in that it puts the module body in the context of the require. So

```
(require a)
code1
(require* b)
code2
```

is like

```
(module-begin
 (require a)
 code1
 (do-require* b
  (require*-module-begin code2)))
```

The idea being that `require*-module-begin` comes from `b`, so it can use it to customize the rest of the module's expansion, such as by expanding `#%app` into the local app versus the app inside the definition of `b`

soegaard:
> A few regular articles here and there pointing out new and useful packages could go a long way to building Racket's community, since more and more interesting developments in Racket are occurring outside the main distribution.

This is an excellent initiative.

Alexis has a point: it is difficult to get make #%app composable, but used locally it can be very handy.

In the category of "just a small hack to show how #%app works" there is 
[brevity.rkt](https://github.com/soegaard/this-and-that/blob/master/brevity/brevity.rkt)
shows how make a shorthand for indexing into strings, lists etc.

In the category of restricting application to a single argument:
[lambda.rkt](https://github.com/samth/polyconf-2015/blob/bbffd13cb15be3d35ca504494928d9d6102ed208/polyconf/lambda.rkt)

Partial application from Formica: [partial-app.rkt](https://github.com/samsergey/formica/blob/master/private/syntax/partial-app.rkt)

mbutterick:
> Module languages should provide their own #%app if they need custom function application behavior. Libraries should not.

How does this bear on the PR? Are you saying the post shouldn’t be published?

lexi-lambda:
> How does this bear on the PR? Are you saying the post shouldn’t be published?

No. I *am* saying that a note should be added that `#%app` is intended to be defined by module languages, and I’m not sure that `fancy-app` or `compose-app` are good examples to point to. Rackjure is a good one, though, and Rosette and Turnstile are also great. I could be in the minority here, though, so maybe tons of other people think `fancy-app` is the best thing since sliced bread.

jackfirth:
This got a lot more discussion than I anticipated, which is definitely not a bad thing! I'll add some edits to address these comments later tonight.

mbutterick:
> I am saying that a note should be added

That seems like overkill, unless Jack’s work will literally cause readers to spontaneously combust. :fire:

> maybe tons of other people think fancy-app is the best thing 

[At least one person](https://github.com/racket/racket-lang-org/pull/33#issuecomment-269723745) seems to like `fancy-app`.

lexi-lambda:
> That seems like overkill, unless Jack’s work will literally cause readers to spontaneously combust.

I’m not sure I understand why a request to add a note of clarification may only be motivated by the possibility of spontaneous combustion.

jackfirth:
I agree with much of Alexis's concern but elided mention of these subtleties in a first draft in the interest of getting a PR up quickly.

blog/_src/posts/2017-02-05-on-application.md
```diff
@@ -0,0 +1,108 @@
+    Title: On Application
+    Date: 2017-02-05T15:17:41
+    Tags:
+
+*posted by Jack Firth*
+
+Today I wanted to write about function application. Specifically, how to redefine and customize
+application with Racket. We'll also look at some Racket packages the define interesting and useful
```
bennn: "the" -> "that"

blog/_src/posts/2017-02-05-on-application.md
```diff
@@ -0,0 +1,108 @@
...
+forms of function application.
+
+## The theory
+
+Application defines one half of *lambda calculus*, the formal model underlying much of modern
+functional programming. The other half is *abstraction*, which creates new functions. Creating and
```
bennn: "which is creating new functions" ?

blog/_src/posts/2017-02-05-on-application.md
```diff
@@ -0,0 +1,108 @@
...
+applying functions lies at the heart of Racket and many other functional languages.
+
+So how are functions applied in Racket? What makes `(if (< 5 10) 'foo 'bar)` a macro use and
+`(< 5 10)` a function use?
+
+That's actually a trick question, because function application *is* a macro in Racket. During macro
+expansion, the Racket expander inspects the first element of an expression to determine if it has a
```
bennn: add link to "Expansion Steps" in docs
http://docs.racket-lang.org/reference/syntax-model.html#%28part._expand-steps%29

blog/_src/posts/2017-02-05-on-application.md
```diff
@@ -0,0 +1,108 @@
...
+case, the function application is converted to a lambda that takes as input one argument for each
+underscore. For example, `(format "Hello ~a" _)` is equivalent to
+`(lambda (v) (format "Hello ~a" v))`. This is especially useful for whipping up quick lambdas as
+arguments to functions like `map` and `filter`:
+
+```racket
+> (require fancy-app)
+> (map (format "Hello ~a, how are you today?" _)
+       (list "Alice" "Bob" "Eve"))
+'("Hello Alice, how are you today?"
```
bennn: Want to change this to "Hello Adele, can you hear me?" ?
(I guess, Adele Paul .... can't think of a third)

blog/_src/posts/2017-02-05-on-application.md
```diff
@@ -0,0 +1,108 @@
...
+  "Hello Bob, how are you today?"
+  "Hello Eve, how are you today?")
+```
+
+The `rackjure`[racket] package redefines `#%app` to make working with nested dictionaries easier.
+Dictionaries can be used to get and set values for keys when used as procedures, and when
+dictionaries are the *second* value of a function application the first value is interpreted as a
+key and the dictionary's associated value is looked up. See [the documentation][2] for details.
```
bennn: "the documentation" -> "the `rackjure` documentation"

blog/_src/posts/2017-02-05-on-application.md
jackfirth: I'm fine with the way it is now, but if anyone has a particularly witty suggestion with three names that they feel strongly about I'm open to it.

mfelleisen:
> On Feb 7, 2017, at 9:54 PM, Jack Firth <notifications@github.com> wrote:
> 
> @jackfirth commented on this pull request.
> 
> In blog/_src/posts/2017-02-05-on-application.md <https://github.com/racket/racket-lang-org/pull/48>:
> 
> > +## The practice
> +
> +Redefinition of `#%app`[racket] is occasionally used when defining new languages, but a much more
> +pedestrian use is to add some notational shorthand to make certain constructs more convenient. For
> +example, consider the [`fancy-app`][1] package. This package provides an `#%app`[racket] macro that
> +behaves just like normal function application *unless* one or more underscores are used. In that
> +case, the function application is converted to a lambda that takes as input one argument for each
> +underscore. For example, `(format "Hello ~a" _)` is equivalent to
> +`(lambda (v) (format "Hello ~a" v))`. This is especially useful for whipping up quick lambdas as
> +arguments to functions like `map` and `filter`:
> +
> +```racket
> +> (require fancy-app)
> +> (map (format "Hello ~a, how are you today?" _)
> +       (list "Alice" "Bob" "Eve"))
> +'("Hello Alice, how are you today?"
> I'm fine with the way it is now, but if anyone has a particularly witty suggestion with three names that they feel strongly about I'm open to it
> 

Matthew, Robby, Jay.

stamourv:
Merged and published, thanks!

PR #53, opened 2017-03-03
say 'python' in the 'pygments not installed' error message

bennn:
It's not much, but it's a hint that the reason your `all.rkt`
raised an exception is due to the `python` system command and
nothing to do with your Racket


stamourv:
LGTM

PR #54, opened 2017-03-04
blog: languages as dotfiles

bennn:
Encouraging others to make a quick `#lang` with their favorite set of `require`s. Also the `@` reader if you're into that.

- - -
The post advertises:
- `raco pkg new`
- `reprovide`
- `remix`

Please let us know if you think of other good shout-outs.

- - -

Also we tried to be careful about terminology; please comment if we used words with technical meaning too imprecisely.

- - -

Finally we're not happy about some of the formatting. In the picture below:
- why are the numbers off the left margin?
- why is `$ mkdir ....` indented more than the other shell commands?
- why is there so much space above a codeblock and so little below?

Is there anything we can do _just in this post_ to correct the formatting?

screen shot 2017-03-03 at 21 51 21


blog/css/main.css.pp
```diff
@@ -161,33 +161,32 @@
   width: 50%;
 }
 
-pre, div.pygments {
+pre, div.pygments, .SCodeFlow {
```
bennn: The changes in this file are progress at getting code to look right in Scribble. Happy to remove this work-in-progress from the PR if you want.

blog/_src/posts/2017-03-03-languages-as-dotfiles.scrbl
```diff
@@ -0,0 +1,291 @@
...
+        return 0;
+      }
+    }
+}|]
+
+@margin-note{
+  At this point, running @exec{$ raco setup --check-pkg-deps scratch} will
+  report an undeclared dependency on @racket{at-exp-lib}.
+  Make sure to add @racket{at-exp-lib} to the @tt{deps} list in your
+  @tt{info.rkt} file.}
```
jackfirth: Mentioning `--fix-pkg-deps` would be useful here.

bennn:
@LeifAndersen what do you think about posting a "request for comments" on Slack?

LeifAndersen:
If you want. I would be weary of avoiding bikeshedding though.

bennn:
Good point, i think not then. We can always update this post.

bennn:
@stamourv can we publish this post?

blog/css/main.css.pp
stamourv: Yeah, probably best to keep separate.

stamourv:
LGTM modulo comment.

LeifAndersen:
Merged.

Issue #37, opened 2016-12-29
change hover boxes to display:none


mbutterick:
So they get excluded from in-page searching.

lexi-lambda:
Would this make the homepage inaccessible by screen readers?

mbutterick:
Since @LeifAndersen raised the issue, I’ll let @LeifAndersen decide whether hiding the text from searches is more annoying than the current whack-a-mole.

LeifAndersen:
@lexi-lambda brings up a good point. I'll give it a shot and see.

Thanks.

PR #56, opened 2017-05-03
RacketCon 2017 web page

mbutterick:


mbutterick:
This PR makes the main RacketCon page look like this: http://mbtype.com/2017/index.html

bennn:
This is great.

Can you add a "stop" button?
Or have non-wavy text appear on mouse hover?

rfindler:
Triiiiipy! I like it.

(The 2016 link on the bottom is broken.)

Robby

mbutterick:
@rfindler: The RC 2016 link will work after the site is rebuilt, and the 2016 site is moved into its new archival folder.

mbutterick:
@bennn: I’ll look into it, as they say these days

LeifAndersen:
FYI, this website breaks screen readers.

I mean, they can still work, but it basically repeats the text over and over and over again before you get to any content. ;)

LeifAndersen:
So, here's a thought, and I won't be offended if you completely reject it.

I think you can keep the cool floaty effect while also making it easier to read for people who use screen readers. Namely:

keep the floating text, but maybe only have one copy of each word floating around, rather than 3-4 copies. That will make the text look much clearer, still look cool (you could even have it change colors if you want), but also not become a garbled mess for people without good vision.

Just a thought anyway.

rfindler:
Is there a way to tell the screen readers that "these 4 copies of that text are to be ignored" that could make screen readers work will still preserving the awesomeness of the current design?

PS: I showed this to our kids and they loved it. They were mesmerized and then said "wow, that's cool. How did he make that?".

LeifAndersen:
Honestly, I have no idea.

Although I can say that as is, the text is kind of hard on my eyes.

mbutterick:
The screen-reader issue was also raised in issue #37, with no resolution. As before, I’m willing to make adjustments — for most browser-layout issues, there’s some kind of workaround — if someone wants to make a PR, tell me what the workaround should be, or just point me to current information about tailoring HTML/CSS to screen readers.

mbutterick:
> They were mesmerized and then said "wow, that's cool. How did he make that?"

@rfindler Maybe it’s time to have the Racket Talk with your kids.

lexi-lambda:
@mbutterick The relevant HTML attributes for accessibility are defined by [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA), which provides control over how screen readers and other devices interpret web pages. In this case, I think you could just add `aria-hidden="true"` to the duplicate elements to help screen readers cope.

mbutterick:
Thanks. That’s done. 67358b2b7caa3529aa0f053d4b92e2e1e98b4783

stamourv:
It's live!

bennn:
@mbutterick I just read this quote on Eelco Visser's [homepage](http://eelcovisser.org/) and thought of you:

> MetaBorg provides generic technology for allowing a host language (collective) to incorporate and assimilate external domains (cultures) in order to strengthen itself. The ease of implementing embeddings makes resistance futile.

PR #57, opened 2017-05-12
Change module location to match new version of frog.

LeifAndersen:
New versions of frog have moved the following two modules to new locations:

frog/xexpr2text -> frog/private/xexpr2text
frog/bodies-page -> frog/private/bodies-page

I have moved these so that the repo builds. Although in general
it would make sense to stop relying on them, or convert frog to have
public versions of the functionality.

### DIFF
--- a/blog/_src/xml-from-blogger/xml.rkt
+++ b/blog/_src/xml-from-blogger/xml.rkt
@@ -1,5 +1,5 @@
 #lang at-exp racket
-(require xml txexpr pollen/unstable/convert pollen/core gregor frog/xexpr2text frog/bodies-page racket/runtime-path)
+(require xml txexpr pollen/unstable/convert pollen/core gregor frog/private/xexpr2text frog/private/bodies-page racket/runtime-path)
 (provide write-mds)
 
 (define-runtime-path blog-xml-file "blog.xml")


samth:
@greghendershott Is breaking things like this intended?

LeifAndersen:
I should note I did open an issue about this in the frog repo: https://github.com/greghendershott/frog/issues/195

greghendershott:
@samth I [commented on the frog issue](https://github.com/greghendershott/frog/issues/195#issuecomment-301213467) before I noticed the issue here.

I didn't realize Frog was being used as a library:

> Ouch. Sorry. Until this recent change, where users' `frog.rkt` can `require` certain Frog modules, I considered Frog to be an application -- not a library. So I _thought_ I was being very responsible to sort out public vs. private, now. I didn't realize some people thought the whole thing was public, already. 😞 
>
> ISTR `xexpr2txt` as some ad hoc code that just (barely) does what Frog needs. Not something I'd offer as a great general solution for xexprs -> plain text. But let me take another look....  I'd also be totally cool if you wanted to copy-pasta...?  Or make a new Racket package for this that can grow in whatever ways people might want....?

greghendershott:
Also, although I've only looked briefly at your code, the internal Frog functions [only seem to be used as part of converting from the old Blogger](https://github.com/racket/racket-lang-org/blob/a45716a199099c5ef5cacb156e92c03cfce175f8/blog/_src/xml-from-blogger/xml.rkt)?

Honestly I'd be fine if you copy-pasta my two `.rkt` files in your directory. The code isn't part of some API to use Frog on an ongoing basis.

greghendershott:
And another also :smile: -- I didn't lob this change out suddenly. I made [a PR against my own repo](https://github.com/greghendershott/frog/pull/194), kept it open for several weeks, and there was a nearly 50 comment discussion on it.

I really tried to do the right thing here. 😞

mbutterick:
> the internal Frog functions only seem to be used as part of converting from the old Blogger?

Correct. The idea was to convert the XML dump of the blog into Frog-compatible Markdown. AFAICT I only ended up needing `xexpr->markdown` from `frog/xexpr2text`. That whole `xml-from-blogger` folder contains files that I retained for archival purposes, since it's the only record of the former blog. It is not used to build the current blog.

mbutterick:
> Is breaking things like this intended?

Greg is blameless. This was strictly an off-label use of `frog`

LeifAndersen:
I agree with @mbutterick here.

Still, do we want to merge this PR?

PR #59, opened 2017-10-12
add post "Tutorial: Creating a Package"

stchang:
[Rendered](https://github.com/stchang/racket-lang-org/blob/create-pkg-tutorial/blog/_src/posts/2017-10-11-tutorial-creating-a-package.md)

### DIFF
--- a/blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
+++ b/blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
@@ -272,16 +272,31 @@ catalog, and then download and install the package source files from that repo.
 
 The package we created in this tutorial consists of a single
 collection. Alternatively, packages may contain several modules spanning
-multiple collections. For example, a package might want to add a module to both
-the `data` and `file` collections.
+multiple collections.
 
 Such packages must be declared as multi-collection packages by changing the
 `collection` entry in `info.rkt` to have value `'multi`. This directs Racket
 package installation to treat each subdirectory in the package as its own
-collection (or partial collection). See the [documentation][doc-multipkg] for
-more details.
+collection (or partial collection).
+
+For example, the [`drracket` package][repo-drracket] leverages this
+organization style to implement its toolbar. Specifically, it adds a `tool.rkt`
+module to many different collections, such as the `macro-debugger` and
+`scribble`, in order to access their callback hooks. Observe that in addition
+to the package's root `info.rkt`, *each* collection in this kind of package
+uses its own `info.rkt` to specify collection-specific information such as
+documentation.
+
+Indeed, any package that wishes to add to an existing collection, even if it's
+just one collection, should be declared as `'multi`. See the
+[`persistent-array` package][repo-persist] for an example.
+
+[repo-drracket]: https://github.com/racket/drracket/tree/master/drracket
+[repo-persist]: https://github.com/samth/persistent-array
 
 ## A Final Note on Multi-Package Libraries
+
+**NOTE:** Most programmers will not need to worry about this section.
 
 If you've browsed the Racket source files at all, you may have noticed that
 many core libraries further subdivide their contents into *several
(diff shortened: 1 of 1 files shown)


blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
```diff
@@ -0,0 +1,284 @@
...
+github repos or even local directories may act as packages as well.
+
+[doc-module]: http://docs.racket-lang.org/guide/module-basics.html
+[doc-collect]: http://docs.racket-lang.org/guide/module-basics.html#%28tech._collection%29
+[doc-package]: http://docs.racket-lang.org/pkg/Package_Concepts.html?#%28tech._package%29
+
+## Local Package Installation
+
+It's best to follow this tutorial with a running example. We'll use [this
+one][repo].
```
bennn: Maybe good to say: "clone the repo & cd into it"

blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
```diff
@@ -0,0 +1,284 @@
...
+Running `raco pkg install` as described above installs the contents of the
+current directory as a single-collection package, using the directory name as
+the name of both the package and the collection.
+
+Alternatively, we can give `raco pkg install` an explicit path. The following
+command, executed from the parent directory, is identical to the above command:
+
+```bash
+# run from parent of `plot-bestfit/` directory
+# you may need to run `raco pkg remove plot-bestfit` before trying this command
```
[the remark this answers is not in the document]
bennn: oh wow ... `raco pkg install plot-bestfi` will succeed no matter what. That's bad ... is it too late to rename the example repo?

blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
```diff
@@ -0,0 +1,284 @@
...
+;; info.rkt
+(define scribblings '(("scribblings/bestfit.scrbl")))
+```
+
+Formally, the [`scribblings` entry][doc-scribblings] is a list of lists, where
+each sublist begins with a documentation source filename and is followed by
+various options. In our `info.rkt` file, we have one sublist that contains only
+the documentation source file and does not specify any other options.
+
+See the [notes from the "Scribbling documentation" tutorial](https://gist.github.com/florence/b3fcc1df922008604e64362484dc1c28) to learn how to write documentation.
```
bennn: I think this section should say how to build and view the documentation.

blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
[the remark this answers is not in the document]
bennn: yes

blog/_src/posts/2017-10-11-tutorial-creating-a-package.md
```diff
@@ -0,0 +1,297 @@
...
+
+In Racket, a library is just a [module][doc-module] (typically in its own file)
+that exports some bindings. A programmer may import such a module using a
+(relative or absolute) filepath:
+
+```racket
+(require "path/to/filename.rkt")
+```
+
+A Racket [**collection**][doc-collect] is a "hierarchical group of modules"
```
stamourv: You may want to link to this post, which discusses this topic in more depth:
http://blog.racket-lang.org/2015/08/modules-packages-and-collections.html

samth:
I think the discussion about single vs multi collection packages confuses two different issues. The are multi-collection packages, which are just about one level of extra directory hirerarchy.

stchang:
Ok, so should both topics be mentioned, just more distinctly? or only one? or neither?

samth:
I would mention multi-collection packages and explain why you would want one, and also mention that the split kind of packages exists but not provide much detail.

samth:
I like these improvements. Two more suggestions:

1. I think you should explicitly say that most people don't need to worry about creating multi-package libraries.
2. I would point to an example of a multi-collection package. I'll see if I can find a good example.

samth:
Here's an example of a multi-collection package: https://github.com/samth/persistent-array

One thing to mention when describing them: there are now _two_ `info.rkt`, one with pkg-level information like dependencies, one with collection-level information like documentation.

rfindler:
There are many multi-collection packages in the standard distribution. Here are some (repo names) from the packages I happen to have installed via --clone: games, 2d, compatibility, data, db, draw, drracket, errortrace, gui, htdp, pict, planet, rackunit, redex, scribble, slideshow, string-constants, syntax-color, webserver, wxme

stchang:
I looked at hundreds of packages to try to find a `'multi` package that actually spanned two separate collections but I could not find one. 

Although I realized that an important point I should mention is that a package must be `'multi` if it wants to add to a known existing collection.

rfindler:
DrRacket does that?

stchang:
Yes, you're right thanks. I overlooked it because I didnt want to conflate `'multi` collections with lib-doc-test repos as @samth pointed out. But I think `drracket` the collection works as a fine example.

rfindler:
gui-lib ?

rfindler:
Beyond gui-lib and drracket, here are three more that might be of interest: htdp-lib, draw-lib and pict-lib.

stamourv:
Merged.

PR #61, opened 2017-10-22
add link to 'racket world map'

bennn:
I'm surprised how much attention the map I posted to google groups is getting, so probably we should put it in a more maintainable place.

Here's one idea. Just link to the google drive map on the homepage.

### DIFF
--- a/www/index.html.pm
+++ b/www/index.html.pm
@@ -222,6 +222,8 @@ Community
 
 ◊link["https://botbot.me/freenode/racket/"]{#racket IRC on freenode.net}
 
+◊link["https://drive.google.com/open?id=1i3zN11e_6te5ytduAiv1cidrIi4&usp=sharing"]{World Map}
+
 ◊link["https://racket.slack.com/"]{Slack channel} (visit ◊link["http://racket-slack.herokuapp.com/"]{this link} to sign up)
 
 ◊link["https://twitter.com/racketlang"]{@racketlang on Twitter}


bennn:
The map is on the wiki: https://github.com/racket/racket/wiki

stamourv:
I think it would be nice to have it on the web page. Most people won't think to check the wiki. It's a good place to putting info for linking to, but not very good for discoverability.
@mbutterick: Any thoughts?

mbutterick:
Maybe at the bottom of [this page](http://racket-lang.org/plt.html)? [[Source](https://github.com/racket/racket-lang-org/blob/master/www/plt.html.pm)]

If you look at the [RacketCon home page](https://github.com/racket/racket-lang-org/blob/master/rcon/2017/index.html.pm) you can see how to embed a Google Map (with a custom color scheme, even).

rfindler:
How about the top? (in its own section)

Robby


On Sun, Oct 22, 2017 at 4:53 PM, Matthew Butterick <notifications@github.com

stamourv:
I'm not sure that page is a good home. It's about more about PLT (the research group) than about users.
Would another (sibling) page make sense?

samth:
I think we should make that page more about the Racket community and contributors anyway. This link would be a good start.

rfindler:
Let’s embed the map.

mbutterick:
Drop pins for the listed PLT shareholders & research centers too.

mbutterick:
(Unless they are in witness protection programs, etc.)

PR #62, opened 2017-10-23
Clarify the plt page some.

samth:
This PR removes the focus on PLT Design Inc, and changes the positioning a bit wrt the research aspects of the project.

### DIFF
--- a/www/plt.html.pm
+++ b/www/plt.html.pm
@@ -15,9 +15,9 @@
 
 
 ◊section{
-◊div{PLT shareholders
+◊div{Recognized Contributors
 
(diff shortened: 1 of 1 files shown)


stamourv:
I don't like "recognized contributors". Are the others unrecognized?
That list is not really representative of Racket contributors anyway.

bennn:
I like the title "Recognized Contributors", but I thinks it's important to say these people are recognized as shareholders in PLT design inc.

Two suggestion: (1) keep the original text, or (2) add a footnote / box that appears on hover to clarify what "special recognition" means.

samth:
The goal here is to not make being a shareholder in a private company be an important part of contribution to Racket. Happy to take suggestions for other phrasings either for the section heading or for the text.

stamourv:
Merged.

PR #65, opened 2017-11-04
initial refinement post draft

pnwamk:
First draft of a post briefly describing working with refinement and dependent function types in Typed Racket.

Any thoughts/feedback for improvement would be great!

Sorry if the format for adding a post to the blog is wrong. I made a quick guess--will do more to fix if/when needed =)

### DIFF
--- a/blog/_src/posts/2017-11-11-adding-refinement-types.md
+++ b/blog/_src/posts/2017-11-11-adding-refinement-types.md
@@ -1,0 +1,540 @@
+    Title: Refinement Types in Typed Racket
+    Date: 2017-11-11T09:00:00
+    Tags:
+
+*posted by Andrew Kent*
+
+With the Racket v6.11 release, Typed Racket has begun to support
+basic refinement and dependent function types.
+This post gives an overview for working with these types while writing
+some simple vector operations.
+
+Currently these types are documented under Typed Racket's
+[experimental
+features](https://docs.racket-lang.org/ts-reference/Experimental_Features.html). They
+are experimental in the sense that they are relatively new. We look
+forward to improving their usefulness by incorporating user feedback.
+
+**Note:** If you wish to experiment with refinement and/or
+dependent function types in the near future, you should use a [snapshot
+build](https://pre.racket-lang.org/installers/), as it will have
+additional features and bug fixes from user feedback. Some of the
+examples in this post will require a build from early November 2017
+or later.
+
+<!-- more -->
+
+
+# Refinement Types
+
+Let us examine the type of `vector-ref` in Typed Racket:
+
+```racket
+(All (a) (-> (Vectorof a) Integer a))
+```
+
+It states that the index (the second argument) must be an `Integer`,
+i.e. any whole-number value in the exclusive range (-∞ , ∞). This
+description, however, is somewhat coarse: we obviously cannot use
+_any_ integer. To avoid a runtime error when indexing into a vector
+`v`, we must use an integer in the range [0, `(vector-length
+v)`). With a refinement type we can precisely describe this subset of
+the integers:
+
+```racket
+(Refine [i : Integer] (<= 0 i (- (vector-length v) 1)))
+```
+
+This refinement type describes values `i` of type `Integer` such that
+`(<= 0 i (- (vector-length v) 1))` produces `#true`. The documentation
+for [Refine](http://docs.racket-lang.org/search/index.html?q=Refine)
+describes precisely what propositions can be used in
+refinements. Today it includes logical statements about the types of
+terms, standard logical combinators (e.g. and, or), and linear integer
+constraints.
+
+# Dependent function types
+
+By combining refinement types with dependent function types
+(i.e. functions whose domain and range may depend on argument and/or
+program values), we can now give a more precise function type for many
+basic vector operations, e.g.:
+
+```racket
+(: safe-vector-ref 
+   (All (A) (-> ([v : (Vectorof A)]
+                 [n : (v) (Refine [i : Integer]
+                            (<= 0 i (- (vector-length v) 1)))])
+                A)))
+(define safe-vector-ref vector-ref)
+```
+
+Here we have stated that `safe-vector-ref`'s second argument (`n`)
+depends on its first argument (`v`) and that the type of `n` is an
+`Integer` which is a valid index for `v`.
+
+If we try and use `safe-vector-ref` with a vector and index which
+Typed Racket cannot prove satisfy the respective types:
+
+```racket
+(safe-vector-ref (vector 0 1) 2)
+```
+
+Typed Racket will report an error while type checking:
+
+```
+Type Checker: type mismatch
+  expected: (Refine
+             (i : Integer)
+             (and (<= i (- (vector-length v) 1)) (<= 0 v)))
+  given: (Refine (z : Positive-Byte) (= 2 z)) in: 2
+```
+
+## Preconditions
+
+Sometimes the refinements used in a dependent function type can also
+be written conveniently as a "precondition", e.g.:
+
+```racket
+(: safe-vector-ref
+   (All (A) (-> ([v : (Vectorof A)]
+                 [n : Integer])
+                #:pre (v n)
+                (<= 0 n (- (vector-length v) 1))
+                A)))
+(define safe-vector-ref vector-ref)
+```
+
+This function type is equivalent to the previous version, but the
+refining proposition that requires that `n` be a valid index for `v`
+now references `n` directly and has been written after the
+list of arguments:
+
+```racket
+#:pre (v n) (<= 0 n (- (vector-length v) 1))
+```
+
+We can read the precondition specification above as follows: the
+precondition for the function depends on arguments `v` and `n` and it
+requires that Typed Racket can prove `(<= 0 n (- (vector-length v)
+1))` in order for a call to be well-typed.
+
+Typed Racket may be able to provide more detailed feedback when the
+arguments are of the correct type but the precondition cannot be
+proved:
+
+```
+Type Checker: could not apply function;
+ unable to prove precondition
+  precondition: (<= 2 (- (vector-length v) 1)) 
+  in: (safe-vector-ref (vector 0 1) 2)
+```
+
+Here, Typed Racket first performed type inference and confirmed that
+the arguments `(vector 0 1)` and `2` had acceptable types before
+discovering the precondition was unprovable at the call site.
+
+We believe the error messages produced when using preconditions may be
+more informative in general. Your mileage may vary.
+
+# Programming with refinement types
+
+_Note: the below examples will work on Racket v6.11 with one
+additional snippet of [code](#makevectortype) or on any Racket build
+more recent than November 11 2017._
+
+When programming with refinement types, use the `#:with-refinements`
+option on the `#lang` line to tell Typed Racket to track additional
+logical information about program terms:
+
+```racket
+;; safe-vector-ops.rkt
+#lang typed/racket/base #:with-refinements
+```
+
+In particular, this will tell Typed Racket to remember that the vector
+produced by `(make-vector n ...)` has length `n`, which will be
+essential for typing our vector operations.
+
+Now we can also declare more detailed types for `vector-set!` and
+`vector-ref` since these more specific types are subtypes of their
+more permissive standard types. If we're feeling adventurous, we may
+use the `unsafe-` variants from `racket/unsafe/ops` that do not check
+their bounds, since Typed Racket proves they are never used
+with invalid indices:
+
+```racket
+(require (only-in racket/unsafe/ops
+                  unsafe-vector-ref
+                  unsafe-vector-set!))
+
+(: safe-vector-ref
+   (All (A) (-> ([v : (Vectorof A)]
+                 [n : Natural])
+                #:pre (v n)
+                (< n (vector-length v))
+                A)))
+(define safe-vector-ref unsafe-vector-ref)
+
+(: safe-vector-set!
+   (All (A) (-> ([v : (Vectorof A)]
+                 [n : Natural]
+                 [a : A])
+                #:pre (v n)
+                (< n (vector-length v))
+                Void)))
+(define safe-vector-set! unsafe-vector-set!)
+```
+
+Now that we have these foundational operations in place
+(i.e. `make-vector`, `safe-vector-ref`, and `safe-vector-set!`) ,
+let's see if we can define some of the [vector
+operations](https://docs.racket-lang.org/reference/vectors.html) that
+are provided by `racket/base` and `racket/vector`, but with more
+precise types!
+
+## `build-vector`
+
+Let's look at the documentation for `build-vector`:
+
+```
+(build-vector n proc) → vector?
+  n : exact-nonnegative-integer?
+  proc : (exact-nonnegative-integer? . -> . any/c)
+
+Creates a vector of n elements by applying proc to the
+integers from 0 to (sub1 n) in order. If vec is the 
+resulting vector, then (vector-ref vec i) is the value 
+produced by (proc i).
+
+Example:
+> (build-vector 5 add1)
+'#(1 2 3 4 5)
+```
+
+First let's consider its type signature. It seems like there are two
+places where we could use refinement types to create a more precise
+type-based specification:
+
+1. the `proc` procedure will always be given an integer in the range
+[0, `n`), and
+
+2. the length of the returned vector is equal to `n`.
+
+We can encode both of these directly in the type of `build-vector`:
+
+```racket
+(: build-vector
+   (All (A) (-> ([n : Natural]
+                 [proc : (n)
+                   (-> (Refine [i : Natural] (< i n))
+                       A)])
+                (Refine [v : (Vectorof A)]
+                  (= n (vector-length v))))))
+```
+
+With the type signature in place, we can proceed to define the
+function:
+
+```racket
+(define (build-vector n proc)
+  (cond
+    [(> n 0)
+     (define init-val (proc 0))
+     (define vec (make-vector n init-val))
+     (let loop! ([i : Natural 1])
+       (when (< i n)
+         (safe-vector-set! vec i (proc i))
+         (loop! (add1 i))))
+     vec]
+    [else (vector)]))
+```
+
+There are a few things to note about the definition:
+
+a. In order to make a `(Vectorof A)`, we need some value of type `A`
+to initialize the vector with (i.e. the second arg to
+`make-vector`). Because of this, we first check if `n` is greater than
+0. If it is greater than 0, we can compute `(proc 0)` and use that to
+initialize the vector we create, otherwise we simply return `(vector)`
+(i.e. an empty vector) which is "vacuously" a vector of `A` for any
+`A`.
+
+b. The recursive loop we use to populate indices 1 through `(sub1 n)`,
+which we named `loop!`, requires a type annotation `Natural` on its
+argument `i` so that its being greater than or equal to 0 is known in
+the body of the recursive function. That, combined with the test `(< i
+n)`, allows us to successfully use `safe-vector-set!` to update the
+contents at `i`. (Note: if Typed Racket had smarter type inference,
+this annotation may not be necessary.)
+
+
+## `vector-map`
+
+`vector-map` takes a vector `vec` and a function `proc` and produces a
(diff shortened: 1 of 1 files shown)


blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+
+```
+Type Checker: type mismatch
+  expected: (Refine
+             (i : Integer)
+             (and (<= i (- (vector-length v) 1)) (<= 0 v)))
+  given: (Refine (z : Positive-Byte) (= 2 z)) in: 2
+```
+
+When a refinement on an argument's type may also be written conveniently as a "precondition", e.g.:
```
dvanhorn: This is a pretty big and unexpected jump.  Should the reader know about `#:pre` and the idea that a refinement may also be formulated as a precondition (I didn't)?  Also, you should explain the precondition first, then the error message, not the reverse.

One thing that's not really addressed later in the document is when should someone use a refinement vs precondition.

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+
+We can read the precondition specification above as follows: the precondition for the function depends are arguments `v` and `n`, and it requires that Typed Racket can prove `(<= 0 n (- (vector-length v) 1))` would produce `#true` at all call sites.
+
+## Programming with refinement types
+
+_Note: these examples should work on any Racket builds from 3 November or later.
+To use with Racket v6.11, an additional small snippet of code (included at the end of
+this post) is required._
+
+When programming with refinement types, use the `#:with-refinements` option on the `#lang` line to tell Typed Racket to more liberally track detailed logical information about program terms:
```
dvanhorn: Using "liberally" here seems correct in the sense of "in generous amounts" like "sprinkle liberally with salt" but it's at odds with "liberally" meaning "in an imprecise or loose manner," which the opposite of what you're trying to convey here.  I'd pick a different word that just has the first sense.

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+Typed Racket may be able to provide more detailed feedback when the arguments are of the correct type but the precondition cannot be proved:
+
+```
+Type Checker: could not apply function;
+ unable to prove precondition
+  precondition: (<= 2 (- (vector-length v) 1)) 
+  in: (safe-vector-ref (vector 0 1) 2)
+```
+
+We can read the precondition specification above as follows: the precondition for the function depends are arguments `v` and `n`, and it requires that Typed Racket can prove `(<= 0 n (- (vector-length v) 1))` would produce `#true` at all call sites.
```
dvanhorn: Something is wrong after the colon.  Should it be "the precondition for the function depends on"?

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+Example:
+> (build-vector 5 add1)
+'#(1 2 3 4 5)
+```
+
+First let's consider its type signature. It seems like there are two places where we could use refinement types to create a more precise type-based specification:
+
+1. the `proc` procedure will always be given an integer in the range [0, n)
+
+2. the length of the returned vector is equal to `n`
```
dvanhorn: Itemization needs punctuation,  e.g.

1. blah blah, and
2. blah blah.

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+         (loop! (add1 i))))
+     vec]
+    [else (vector)]))
+```
+
+There are a few things to note about the definition:
+
+a. In order to make a `(Vectorof A)`, we need some value of type `A` to initialize the vector with (i.e. the second arg to `make-vector`). Because of this, we first check if `n` is  greater than 0. If it is greater than 0, we can compute `(proc 0)` and use that to initialize the vector we create, otherwise we simply return `(vector)` (i.e. an empty vector) which is  "vacuously" a vector of `A` for any `A`.
+
+b. The recursive loop we use to populate indices 1 through `(sub1 n)`, which we named `loop!`, requires a type annotation `Natural` on its argument `i` so that its being greater than or equal to 0 is known in the body of the recursive function. That, combined with the test `(< i n)`, allows us to successfully use `safe-vector-set!` to update the contents at `i`. (Note: if Typed Racket's type inference were not simple local type inference, this annotation may not be necessary)
```
dvanhorn: Last parenthetical remark needs punctuation.  (Blah blah.)

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+     (define res (make-vector len (safe-vector-ref vec start)))
+     (let loop! ([i : Natural 0])
+       (when (< i len)
+         (define a (safe-vector-ref vec (+ start i)))
+         (safe-vector-set! res i a)
+         (loop! (+ 1 i))))
+     res]))
+```
+
+Using a precondition we can assert the range specified by `start` and `end` is some sensible (although possibly empty) portion of the vector `vec`. A `Refine` allows us to indicate that the length of the returned vector will always be `(- end start)`.
```
dvanhorn: "Using a precondition,".

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Dependent function type limitations
+
+Dependent function types currently only support mandatory positional arguments. In the future we plan to add optional, rest, and keyword argument support.
+
+Also, currently functions with dependencies between arguments and/or preconditions cannot be used within a `case->` form.
+
+### Bugs
+
+If you have found a bug or have an idea for how these features might be more useful, feel free to open an issue on Typed Racket's [github repo](https://github.com/racket/typed-racket/issues).
```
dvanhorn: What's the story for untyped modules that interact with typed modules that use refinement types?

pnwamk:
Thanks for the feedback! Will try to incorporate it all later tonight.

pnwamk:
Update: I've attempted to address the issues @dvanhorn pointed out and fix some other typos and formatting.

stamourv:
Just let me know when it's ready to merge.

FWIW, I think the examples are not very exciting.

samth:
@stamourv Do you have suggestions for more exciting examples? Or categories of examples that you think would be more exciting?

stamourv:
Not concretely.
Maybe something dealing with a file format / protocol that has field length information as part of the format. I guess maybe like TCP payload size? ASN.1 also has stuff like that, I think, but that's not very friendly. Possibly some media codecs may have similar features.

samth:
That seems like it's unlikely to work with TR currently -- we don't have dependent struct fields.

dvanhorn:
Matrix operations?

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+
+In particular, this will tell Typed Racket to remember that the vector
+produced by `(make-vector n ...)` has length `n`, which will be
+essential for typing our vector operations.
+
+Now we can also declare more detailed types for `vector-set!` and
+`vector-ref` since these more specific types are subtypes of their
+more permissive standard types. If we're feeling adventurous, we may
+use the `unsafe-` variants from `racket/unsafe/ops` that do not check
+their bounds, since Typed Racket will be proving they are never used
```
samth: Typed Racket proves

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+initialize the vector we create, otherwise we simply return `(vector)`
+(i.e. an empty vector) which is "vacuously" a vector of `A` for any
+`A`.
+
+b. The recursive loop we use to populate indices 1 through `(sub1 n)`,
+which we named `loop!`, requires a type annotation `Natural` on its
+argument `i` so that its being greater than or equal to 0 is known in
+the body of the recursive function. That, combined with the test `(< i
+n)`, allows us to successfully use `safe-vector-set!` to update the
+contents at `i`. (Note: if Typed Racket's type inference were not
```
samth: Note: if Typed Racket's type inference were smarter here, this annotation would not be necessary.

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+  (let loop! ([i : Natural 0])
+    (when (< i count)
+      (define a (safe-vector-ref src (+ src-start i)))
+      (safe-vector-set! dest (+ dest-start i) a)
+      (loop! (+ 1 i)))))
+```
+
+## Typed/Untyped interaction
+
+Typed Racket will generate dependent function contracts (e.g. `->i`
```
samth: Can this link to the `->i` docs?

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,323 @@
...
+   (All (A) (-> ([v : (Vectorof A)]
+                 [n : (v) (Refine [i : Integer]
+                                  (<= 0 i (- (vector-length v) 1)))])
+                A)))
+(define safe-vector-ref vector-ref)
+```
+
+Here we have stated that `safe-vector-ref`'s second argument (`n`) depends on its first argument (`v`) and that the type of `n` is an `Integer` which is a valid index for `v`.
+
+If we try and use `safe-vector-ref` with a vector and index which Typed Racket cannot prove satisfy the respective types:
```
bennn: Do you really want to link to the search page?

Link for `Refine` in typed racket:
http://docs.racket-lang.org/ts-reference/Experimental_Features.html#%28form._%28%28lib._typed-racket%2Fbase-env%2Fbase-types-extra..rkt%29._.Refine%29%29

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,323 @@
...
+
+```racket
+(safe-vector-ref (vector 0 1) 2)
+```
```
bennn: Maybe add: "(e.g. expressions using `<`, `+`, `-`, and `*`)"

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,323 @@
...
+## Programming with refinement types
+
+_Note: these examples should work on any Racket builds from 3 November or later.
+To use with Racket v6.11, an additional small snippet of code (included at the end of
+this post) is required._
+
+When programming with refinement types, use the `#:with-refinements` option on the `#lang` line to tell Typed Racket to track additional logical information about program terms:
+
+```racket
+;; safe-vector-ops.rkt
```
bennn: "directly after" => "after" ?

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,323 @@
...
+#lang typed/racket/base #:with-refinements
+```
+
+In particular, this will tell Typed Racket to remember that the vector produced by `(make-vector n ...)` has length `n`, which will be essential for typing our vector operations.
+
+Now we can also declare more detailed types for `vector-set!` and `vector-ref` since these more specific types are subtypes of their more permissive standard types. If we're feeling adventurous, we may use the `unsafe-` variants from `racket/unsafe/ops` that do not check their bounds, since Typed Racket will be proving they are never used with invalid indices:
+
+```racket
+(require (only-in racket/unsafe/ops unsafe-vector-ref
+                                    unsafe-vector-set!))
```
bennn: Wait, this sounds backwards. "You can write this type if TR can prove something about all call site".

Suggestion:
> the precondition for `safe-vector-ref` depends on arguments `v` and `n` and states that an application of `safe-vector-ref` to a vector `v` and integer `n` is well typed only if `(<= 0 n (- (vector-length v) 1))`

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+
+(safe-vector-ref (vector 0 1) 2)
+
+;; produces:
+;;
+;; safe-vector-ref: contract violation
+;;  #:pre condition violation; ...
+```
+
+As always, care should be taken when designing APIs that involve
```
bennn: suggestion: replace this paragraph with something like "the contract will check the arguments at runtime".

(Better to say _why_ care should be taken than to just say "beware")

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+discovering the precondition was unprovable at the call site.
+
+We believe the error messages produced when using preconditions may be
+more informative in general. Your mileage may vary.
+
+## Programming with refinement types
+
+_Note: these examples should work on any Racket builds from 3 November
+2017 or later.  To use with Racket v6.11, an additional small snippet
+of code (included at the end of this post) is required._
```
bennn: This text should link to that part of the post.

I think you'll want to do that with HTML:
https://stackoverflow.com/a/6494918/5237018

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+list of arguments:
+
+```racket
+#:pre (v n) (<= 0 n (- (vector-length v) 1))
+```
+
+We can read the precondition specification above as follows: the
+precondition for the function depends on arguments `v` and `n` and it
+requires that Typed Racket can prove `(<= 0 n (- (vector-length v)
+1))` would produce `#true` at all call sites.
```
bennn: This sounds backwards to me, like "you can only write this type if TR can prove something about all call sites"

suggestion:
> the precondition for `safe-vector-ref` depends on arguments `v` and `n` and states that a call `(safe-vector-ref v n)` is well typed only if `(<= 0 n (- (vector-length v))`

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+   (All (A) (-> ([v : (Vectorof A)]
+                 [n : Integer])
+                #:pre (v n) (<= 0 n (- (vector-length v) 1))
+                A)))
+(define safe-vector-ref vector-ref)
+```
+
+This function type is equivalent to the previous version, but the
+refining proposition that requires that `n` be a valid index for `v`
+now directly references `n` and has been written directly after the
```
bennn: "directly after" => "after" ?

blog/_src/posts/2017-11-XX-adding-refinement-types.md
```diff
@@ -0,0 +1,449 @@
...
+v)`). With a refinement type we can precisely describe this subset of
+the integers:
+
+```racket
+(Refine [i : Integer] (<= 0 i (- (vector-length v) 1)))
+```
+
+This refinement type describes values `i` of type `Integer` such that
+`(<= 0 i (- (vector-length v) 1))` produces `#true`. The documentation
+for [Refine](http://docs.racket-lang.org/search/index.html?q=Refine)
```
bennn: Do you really want to link to the search page?

Link for `Refine` in typed racket:
http://docs.racket-lang.org/ts-reference/Experimental_Features.html#%28form._%28%28lib._typed-racket%2Fbase-env%2Fbase-types-extra..rkt%29._.Refine%29%29

pnwamk: If Refine moves from the Experimental Features category to somewhere else that link would break -- linking to the search seemed a simple way to "future proof" it against that change.

pnwamk:
As @samth mentioned, some interesting use cases will be easier to express once dependent struct types are supported (e.g. balanced red-black trees).

Would an implementation of quicksort be a somewhat more interesting example?

```racket
(: quicksort! (-> (Vectorof Real) Void))
(define (quicksort! A)
  (quicksort-helper! A 0 (- (vector-length A) 1)))

(: quicksort-helper! (-> ([A : (Vectorof Real)]
                   [lo : Natural]
                   [hi : Integer])
                  #:pre (A lo hi)
                  (and (<= lo (vector-length A))
                       (< hi (vector-length A)))
                  Void))
(define (quicksort-helper! A lo hi)
  (when (< lo hi)
    (define pivot (partition! A lo hi))
    (quicksort-helper! A lo (- pivot 1))
    (quicksort-helper! A (+ pivot 1) hi)))


(: swap! (-> ([A : (Vectorof Real)]
              [i : Natural]
              [j : Natural])
             #:pre (A i j)
             (and (< i (vector-length A))
                  (< j (vector-length A)))
             Void))
(define (swap! A i j)
  (define tmp (safe-vector-ref A i))
  (safe-vector-set! A i (safe-vector-ref A j))
  (safe-vector-set! A j tmp)) 



(: partition! (-> ([A : (Vectorof Real)]
                   [lo : Natural]
                   [hi : Natural])
                  #:pre (A lo hi)
                  (< lo hi (vector-length A))
                  (Refine [pivot : Natural]
                          (<= lo pivot hi))))
(define (partition! A lo hi)
  (define pivot (safe-vector-ref A lo))
  (let outer-loop!
    ([i : (Refine [n : Natural] (< lo n))
        (+ 1 lo)]
     [j : (Refine [n : Natural] (and (<= n hi)
                                     (<= lo n))) hi])
    (let i-loop
      ([i : (Refine [n : Natural] (< lo n))
          i])
      (cond
        [(and (<= i j)
              (< (safe-vector-ref A i) pivot))
         (i-loop (+ i 1))]
        [else
         (let j-loop
           ([j : (Refine [n : Natural] (and (<= n hi)
                                            (<= lo n))) j])
           (cond
             [(and (>= (safe-vector-ref A j) pivot)
                   (>= j i))
              (j-loop (- j 1))]
             [(> i j) (swap! A lo j)
                      j]
             [else
              (swap! A i j)
              (outer-loop! i j)]))]))))
```

pnwamk:
I've added a few more tweaks and the quicksort example. I want to push a quick fix to TR before this is posted (one that the quicksort example relies on), and of course if there are further comments or feedback I'm happy to incorporate that.

Personally, I'm not super concerned with the examples not being "flashy" -- I think they do their job well: they describe the features with thorough examples and illustrate how they can be used today in a way the docs cannot reasonably do. As more type system features are added (i.e. dependent struct fields) and more work is done, I'm sure we'll have some super cute things to show off.

stamourv:
I think quicksort is mildly more interesting.

Re larger point: I don't think it's about being flashy, or showing off. I think it's about motivating why a Racket user or bystander (the main audience of the blog, AIUI) should be interested in this new development. That's something you'd want (to some extent) even in docs. Using toy academic examples works against that, and makes us look out of touch with our users' needs, IMO. This applies not just to blog posts, it's about any form of communication, and ultimately about the way we think about the work we do.

That said, I won't stand in the way of publishing this post.

samth:
@stamourv I think almost every Racket programmer (or non-Racket programmer who might be reading the blog) has worked with vectors at some point. There's nothing "academic" about vectors, nor is making vectors work well showing that we're somehow "out of touch".

The important difference between quicksort and smaller functions is exactly that -- they're small, so it's not clear how it would apply to a real function that a user might write. It's important to show substantial examples, so I'm glad this post does that. But there's nothing wrong with vectors.

stamourv:
Nothing wrong with vectors per se. It's all about what you do with them.
Sorting is more interesting than accessing, but something from some application domain would be more interesting still, IMO.
But I think we're going in circles, so it may be best to stop here.

samth:
I think the code snippets are making the blog too wide currently.

pnwamk:
I think I've fixed the width issue. (I can't be certain though... the various incantations I've tried to render the webpages have not produced .html files that actually use the style/formatting/etc that the racket blog uses... ¯\\\_(ツ)\_/¯)

samth:
I recommend running 'raco frog -bp' in the blog directory.

pnwamk:
Okay - I glanced at it in Safari, Chrome, and Firefox. I think it's good to go.

samth:
https://blog.racket-lang.org/2017/11/adding-refinement-types.html

PR #69, opened 2018-04-24
Make screen readers able to read rcon website.

LeifAndersen:


jeapostrophe:
Nice, that was easier than I thought it would be!

PR #72, opened 2018-06-01
embed the racket world map

bennn:
Hi, I tried to embed a map in the PLT page, as suggested in https://github.com/racket/racket-lang-org/pull/61

This current attempt doesn't seem to work --- I ran `www/all.rkt` and then opened the generated `plt.html` in my browser and got a blank frame (see next comment).

I'm posting this PR in case anyone has suggestions, and so I don't forget to try this again soon!

### DIFF
--- a/www/team.html.pm
+++ b/www/team.html.pm
@@ -14,6 +14,14 @@
 
 
 ◊section{
+◊div{Racket World Map}
+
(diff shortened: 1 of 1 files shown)


bennn:
all set now

screen shot 2018-06-03 at 00 19 12

stamourv:
Merged.

PR #84, opened 2019-01-27
the need to get rid of "Racket is a Scheme".

mfelleisen:
The line is too misleading. Adding "(kind of)" is a compromise. Better ideas welcome.

### DIFF
--- a/www/new-name.rkt
+++ b/www/new-name.rkt
@@ -63,7 +63,7 @@
       arguments, classes, static types, and even curly braces.}
 
 
-    @heading{Racket is a Scheme}
+    @heading{Racket is (kind of) a Scheme}
 
     @nested{Racket is still a dialect of Lisp and a descendant of Scheme.  The
       tools developed by PLT will continue to support R5RS, R6RS, the old


rfindler:
Would it still be subtle if we wrote `Racket is a scheme` (lowercase)?

samth:
What's the audience for this change? Are people currently looking at the "new-name.html" page, even though our web page doesn't link to it? If so, maybe we need a different page that explains things in a less 2010-centric way.

mfelleisen:
I guess I forgot to CC you on the email from the HOPL/POPL meeting. 

Guy Steele googled "Racket Scheme" during the HOPL PC meeting and read this one bold-faced line aloud. Everyone looked at me and basically thought that he had just shown I was a liar when I said Racket isn't a Scheme it is an heir to Scheme (the next line). Nobody seemed to read the next, plain-text line. 

I also reviewed a CUP book proposal for a PL (311). The book is referring to our implementation as Racket Scheme. It uses several language and will show interpreters in both Racket and Python, possibly more. CUP wants to advertise it as a Racket Scheme book.

samth:
The first sounds like a definite problem, and worth clarifying. Could we just use the literal next line: "Racket is a descendant of Scheme"?

The second I see plenty, but I've given up hope than anything we say can fix it generally. :(

mfelleisen:
I'd be happy with this change too. 

As long as none of us are ever exposed to this "Racket is a Scheme" again, I am fine.

PR #87, opened 2019-02-05
RacketCon & Racket School 2019 registration pages

mfelleisen:


mfelleisen:
I am happy with this page for now and to get the ball rolling we should release it.

PR #98, opened 2019-08-12
Racket 7.4 announcement: backquote double dashes to avoid conversion to em-dash

kisaragi-hiu:
Frog (or more specifically, [greghendershott/markdown](https://github.com/greghendershott/markdown)) automatically converts `--` to em-dashes; we don't want this for command-line arguments. Quote them to avoid this behavior.


stamourv:
Excellent idea, thanks!

PR #99, opened 2019-08-14
Try harder to detect platforms.

samth:
See https://groups.google.com/forum/#!topic/racket-users/XAUGE1q793w for background.

### DIFF
--- a/download/download-pages.rkt
+++ b/download/download-pages.rkt
@@ -316,7 +316,8 @@
     function getPlatformOrder() {
       var p = navigator.platform;
       var p2 = navigator.appVersion;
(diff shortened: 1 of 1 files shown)


jbclements:
This looks clear and sensible to me, but I do have one question: Can I ask what the `@";"` is doing? Is that some terrible JS-ism that I should know about?

samth:
It's inserting a real semicolon in the output since this is scribble-generating-js.

jbclements:
so the predicate that succeeds just happens to effectfully generate stdout? Yikes. Ah well.

jbclements:
In the spirit of erring on the side of getting things done, I'm going to go ahead and merge this pull request. Someone please tell me if I'm breaking implicit protocol like I usually do :).

PR #111, opened 2019-12-09
Don't compile pollen.rkt to avoid zo version mismatch errors.

### LINKED ISSUE #110: Build error due to out-of-date version
Every time I re-build Racket with a new VM version, I get an error like this from this package:

```
read-compiled-linklet: version mismatch  expected: "7.5.0.9"  found: "7.5.0.2"  in: /home/samth/sw/plt/extra-pkgs/racket-lang-org/rcon/2019/compiled/ta6le/pollen_rkt.zo
  compilation context...:
   /home/samth/sw/plt/extra-pkgs/racket-lang-org/rcon/2019/text.svg.pm
  context...:
   raise-read-error
```

samth:
Fixes #110.

### DIFF
--- a/rcon/2019/info.rkt
+++ b/rcon/2019/info.rkt
@@ -1,0 +1,3 @@
+#lang info
+
+(define compile-omit-paths '("pollen.rkt"))


samth:
This doesn't actually fix the problem.

PR #105, opened 2019-09-01
Friendly Environment Policy

spdegabrielle:
Adapted from the [RacketCon 2018 Friendly Environment Policy](https://github.com/racket/racket-lang-org/blob/0d8a413ba37bfb1a0bebc7ea8ed5a9da4c76b55b/rcon/2018/index.html.pm#L278) and the [Rust Code of conduct](https://www.rust-lang.org/policies/code-of-conduct) which in turn was adapted from the Node.js Policy on Trolling as well as the Contributor Covenant v1.3.0.

First draft for discussion.

PR #112, opened 2019-12-11
Friendly Environment policy

mflatt:
Continued from #105

### DIFF
--- a/www/friendly.html.pm
+++ b/www/friendly.html.pm
@@ -1,0 +1,26 @@
+#lang pollen
+
+◊top-section{
+◊span[#:id "logo" #:style "font-size:2.3rem;white-space:nowrap;"]{◊link["index.html"]{◊img[#:src "img/racket-logo.svg" #:class "logo"] Racket} ◊span[#:id "tagline" #:class "disappearing" #:style "font-size:70%;color:gray;white-space:nowrap;margin-left:0.2rem;"]{}}
+
+◊div{
+◊link[#:class "top-button disappearing-late" #:id "docs" "https://docs.racket-lang.org/"]{docs}
+
+◊link[#:class "top-button disappearing-late" #:id "packages" "https://pkgs.racket-lang.org/"]{packages}
+
+◊link[#:class "top-button disappearing-late" #:id "download" "https://download.racket-lang.org/"]{download}
+}
+}
+
+
+◊section[#:class "one-column-body-text"]{
+Friendly Environment Policy
+
+The Racket community aims to improve the world through programming. It started with the goal of introducing everyone to the wonderful world of program design, with a spirit of full inclusion and no emphasis on any specific group.  Over time it has grown into a full-fledged professional community with a well-known reputation for helpfulness and openness on its online communication channels.
+
+Racket project management wants to encourage and preserve this open exchange of ideas, which requires an environment that enables all to participate without fear of personal harassment. We define ◊em{harassment} to include specific unacceptable factors and behaviors listed in ◊link["https://www.acm.org/about-acm/policy-against-harassment"]{the ACM's policy aganist harassment}. Unacceptable behavior will not be tolerated.
+
+Anyone witnessing or subject to unacceptable behavior should notify ◊link["team.html#management"]{Racket project management}, which can be reached at ◊link["mailto:management@racket-lang.org"]{management@racket-lang.org}.
+
+◊p[#:id "bottom"]{The wording of this policy is directly derived from ◊link["http://snapl.org/2015/policy.html"]{that of the SNAPL conference}, which is in turn derived from that of the ACM, with thanks to both.}
+}
--- a/www/index.html.pm
+++ b/www/index.html.pm
@@ -231,6 +231,8 @@ Community
 
 ◊link["sfc.html"]{Software Freedom Conservancy}
 
+◊link["friendly.html"]{Friendly Environment policy}
+
 ◊link["sfc.html"]{Donate to Racket}
 Make a tax-deductible contribution to support our work.
 
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ compiled/
 /www/sfc.html
 /www/template.html
 /www/books.html
+/www/friendly.html
 /rcon/2019/template.svg
 /rcon/2019/text.svg
 
--- a/www/css/styles.css.pp
+++ b/www/css/styles.css.pp
@@ -256,7 +256,7 @@ section#logo * {
   padding-bottom: 0;
 }
 
-section#bottom {
+section#bottom, p#bottom {
   font-size: 70%;
 }
 
--- a/www/team.html.pm
+++ b/www/team.html.pm
@@ -56,7 +56,7 @@
 
 ◊link["http://www.deinprogramm.de/sperber/"]{Michael Sperber}
 
-◊link["http://homes.soic.indiana.edu/samth/"]{Sam Tobin-Hochstadt}
+◊link["https://samth.github.io/"]{Sam Tobin-Hochstadt}
 
 ◊link["http://www.ccs.neu.edu/home/ryanc/"]{Ryan Culpepper}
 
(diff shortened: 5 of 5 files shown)


mflatt:
The intent of this revision is to simplify. Instead of setting up a group of moderators, this variant makes explicit that preserving a friendly environment is Racket management's responsibility as well as its goal.

samth:
@lexi-lambda you had some useful comments on the original draft of this, and I'm interested in your thoughts on the revision.

www/index.html.pm
```diff
@@ -231,6 +231,8 @@ Community
 
 ◊link["sfc.html"]{Software Freedom Conservancy}
 
+◊link["friendly.html"]{Friendly Enviornment policy}
```
bennn: typo "Enviornment"

www/friendly.html.pm
```diff
@@ -0,0 +1,26 @@
...
+}
+}
+
+
+◊section[#:class "one-column-body-text"]{
+Friendly Environment Policy
+
+The Racket community aims to improve the world through programming. It started with the goal of introducing everyone to the wonderful world of program design, with a spirit of full inclusion and no emphasis on any specific group.  Over time it has grown into a full-fledged professional community with a well-known reputation for helpfulness and openness on its on-line communication channels. Racket project management wants to encourage and preserve that open exchange of ideas.
+
+This spirit requires an environment that enables all to participate without fear of personal harassment. We define harassment as unwelcome or hostile behavior, that is, behavior that focuses on people instead of ideas. ◊link["http://www.acm.org/sigs/volunteer_resources/officers_manual/anti-harassment-policy"]{The ACM's anti-harassment policy} lists a variety of specific unacceptable factors and behaviors. Racket project management considers responses such as “just joking,” or “teasing,” or being “playful” as unacceptable.
```
lexi-lambda: This link does not go anywhere useful for me (I just get redirected to https://www.acm.org/special-interest-groups/volunteer-resources, which does not have any obvious link to the policy that I could find), which seems like a pretty big problem.

samth: I think it's now at https://www.acm.org/about-acm/policy-against-harassment

mflatt: There's an "ACM Policies" box on the lower right. But that took me a while to find, too, so probably a more direct link is better.

lexi-lambda: Thanks! Since the link seems questionably stable, I think it would be better if the salient points were quoted or paraphrased on this page directly, but maybe it isn’t a big problem.

lexi-lambda: > We define harassment as unwelcome or hostile behavior, that is, behavior that focuses on people instead of ideas.

I would reword this sentence. While I think it means well, as-worded it essentially says “do not criticize someone for their personal beliefs.” Though quite reasonable in a vacuum, this and similar sentiments (such as “be nice” or “be civil”) have sadly been used to take action against people who call attention to discrimination, especially if that discrimination took place in a different forum or community.

I don’t think anyone has any reason to worry about that problem in the Racket community, but those who benefit most from a Code of Conduct don’t have the luxury of already knowing that context. Most people will probably interpret it charitably, but it helps to leave no room for doubt.

I like the way [Rust’s Code of Conduct](https://www.rust-lang.org/policies/code-of-conduct) phrases this:

> We will exclude you from interaction if you insult, demean or harass anyone. […] In particular, we don’t tolerate behavior that excludes people in socially marginalized groups. Private harassment is also unacceptable.

It’s short and to the point, but it’s also very clear.

mfelleisen: I don't like the Rust sentence at all. It is overly aggressive and focused on specific groups. 

We care that every *individual* is treated properly, period. Much stronger, while not being aggressive againt, or focused on, any group.

jackfirth: There is no way to properly address harassment without addressing the fact that some groups of people are far more likely to be harassed than others.

mfelleisen: There is. Respect every individual. Every single one of them. The End.

lexi-lambda: Matthias: I am not suggesting copying the Rust Code of Conduct wholesale. I do suggest incorporating the three sentences I quoted. Rewording them slightly is fine, but muddying their meaning is not.

The first sentence reads, “We will exclude you from interaction if you insult, demean or harass **anyone**.” It is very clear. You may not find the second sentence necessary, but it is hardly offensive, and it will give some people who particularly need it a feeling of additional safety.

With all due respect: don’t die on this particular hill. It doesn’t really affect you. I guarantee it matters quite a lot more to some people than you find it philosophically unnecessary.

mfelleisen: It affects me in very serious ways, and that's precisely why I don't want to speak of goops, ever. As a non-liberal I am routinely attacked---including at a RacketCon a couple of years back. I don't seek help, I don't advertise this, I cope but I ascribe all of this to the group think that is so pervasive in the modern and especially the tech world. 

You have already exclude me from 'users'. If you wish to exclude me from other venues too, this is the way to go.

lexi-lambda: > As a non-liberal I am routinely attacked---including at a RacketCon a couple of years back. I don't seek help, I don't advertise this, I cope but I ascribe all of this to the group think that is so pervasive in the modern and especially the tech world.

If you feel you have been harassed, it seems like this very document provides some recommendations on how that should be handled, with or without my suggestions.

But that is really immaterial. Clearly, you agree that “socially marginalized groups” should be able to participate in Racket community spaces without fear of harassment, seeing as they are a subset of “everyone.” Offering marginalized people—that is, people who are discriminated against more often *by the definition of the word “marginalized”*—extra reassurance that they are, in fact, welcome should not be a political issue. If you feel that sentence is an “attack” on “non-liberals,” I am not sure what to say.

jeapostrophe: I heartily agree with Matthias. We want to foster respect and kindness and love and charity for all people regardless of their membership in any particular groups. No groups deserve more or less respect or protection. By including this sentence, we are favoring some over others.

AlexKnauth: I agree with @lexi-lambda. I agree we should focus on respect for individual people... that's what the first sentence is for. It applies to people whether they're part of an often-marginalized group or not. But that doesn't make the next part unnecessary. I think we should give extra assurance that we don't allow harassment against people in marginalized groups in a follow-up sentence.

mflatt:
I've pushed a revised proposal that removes the attempt to define "harassment" in one sentence and that instead appeals to an existing definition — like the Rust CoC does, but sticking here with the ACM policy. Along those lines, I removed the elaboration about "just joking".

We can't outsource our responsibility to understand and encourage acceptable behavior, but we don't need to innovate in this area. That's why I'm moving the PR further in the direction of relying on the ACM policy.

spdegabrielle:
Thank you @mflatt @samth @jeapostrophe @mfelleisen @lexi-lambda and @AlexKnauth for engaging with this as I believe it is am important issue.

If you are being attacked then it is inevitable that other people **without** the requisite experience and coping skills will also be attacked.

I believe that if a member of the racket management team experiences harassment then they should take action, either by directly excluding the harasser, of if they are worried about conflict of interest, referrer it to another member of the management team to take action. Please tell me if you think I'm wrong as I believe that it the intention of the following line of the FOP; 

>Anyone witnessing or subject to unacceptable behavior should notify ◊link["team.html#management"]{Racket project management}

@lexi-lambda was **invited** to provide feedback, and she kindly took time out oh her day to to provide responses that included concerns, suggested text and explained the rationale. 

Please suggest a text that is clear and you are willing to enforce, because as with the actual instance of harassment above, a policy is pointless if those responsible are not willing to enforce it.

My apologies if I come across as curt or grumpy. This is just what I sound like when I am trying (and probably failing) to write clearly.

Once again, thank you Jay, Matthias and Alexis for taking the time to work on this.

Every day I see examples of the tireless work all three of you do in the Racket community. I believe this is because of your common desire to 'foster respect and kindness and love and charity for all people'.  All three of you have helped me many times.

Thanks again,

Stephen

PS Don't give up!

rmculpepper:
The "just joking" text seems to be reworded but not removed. Was that the intended change?

mflatt:
No, sorry — I pushed the wrong version. It should be as intended, now.

mflatt:
Any objections to the latest version? If not, I plan to merge in a day or so.

www/index.html.pm
jeapostrophe: The typo is still there

jeapostrophe:
I'm good, minus that typo.

mflatt:
Thanks for the reminder – now fixed.

mfelleisen:
Me too.

spdegabrielle:
Just saw the FEP on the website! thank you @mflatt @mfelleisen @jeapostrophe

PR #119, opened 2020-04-26
Partially fixes #67: add an explanation of "Minimal Racket"

### LINKED ISSUE #67: No (OK, very hidden) explanation of "Minimal Racket vs Racket"
On the [download page](https://download.racket-lang.org/), there is an option to download Racket or Minimal Racket, but there's no explanation of what "Minimal Racket" means.  Of the first set of links there: Announcement, Notes, Documentation, More Variants and Checksums, none have any explanation.  Searching the documentation for "Minimal" turns up nothing.

You have to go to "Snapshot Builds", Utah or Northwestern, then click "?" on Minimal Racket to find out what it means: "Includes just enough of Racket that you can use raco pkg to install more."  Then you'd have to navigate back (unless you actually want a snapshot) to the download page.

Seems like there really ought to be a description on the download page.  There probably ought to be the same description on the "More Variants" page as the "Snapshot" page (they have a very similar structure), too.

sorawee:


sorawee:
Is this good to merge? It's been one month, and there are at least three people (at least according to #67), whose confusion would be addressed by this PR.

jbclements:
Yikes... I'm not sure I'm qualified to review this. Have you tested it? I'm going to cross my fingers and merge this.

sorawee:
I tested it, but it would be nice to have someone with a different environment to test it out. But I guess we will see soon how it goes...

PR #121, opened 2020-05-11
Fix serverracket.com not redirecting HTTP->HTTPS

sondr3:
I get 404 on both Firefox and Chromium if I try to access the URL on the frontpage, seemingly because the webpage does not redirect HTTP to HTTPS.

### DIFF
--- a/www/books.html.pm
+++ b/www/books.html.pm
@@ -28,7 +28,7 @@ Learn to program with Racket, one game at a time.
 Make your own programming languages with Racket.
 
 
-◊link["http://serverracket.com"]{Server: Racket}
+◊link["https://serverracket.com"]{Server: Racket}
 Develop a web application with Racket.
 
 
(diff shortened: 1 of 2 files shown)


jessealama:
Thanks! (I'm the guy behind serverracket.com.) I just confirmed that HTTP does redirect to HTTPS, so I'm not sure what's going on here, but thanks for taking care of it.

sondr3:
Weird, I even used DownForEveryone service to check that it wasn’t just me: https://downforeveryoneorjustme.com/serverracket.com

sorawee:
It's not just only you. I can't access it too.

jessealama:
It seems that the load balancer I'm using behind the scenes isn't doing the redirect. I don't know why; I'm on it. (In any case, the HTTPS variant works.)

jessealama:
It should work now. All I did was toggle HTTPS redirection from on to off, then back.

jessealama:
Curiously enough, I used to have some code, in the Racket web application that's behind serverracket.com, for doing HTTP-to-HTTPS redirection. I trashed that bit of code when I found that I could configure the load balancer to do that redirection. I'll resurrect that code.

PR #123, opened 2020-06-01
Fixes #120: revamp JavaScript on the download page

### LINKED ISSUE #120: Buggy download page
1. Navigate to https://download.racket-lang.org/racket-minimal-v7.6.html. Notice that there are two "Variant" select widgets. Only the second one is functional.
2. Click the "Platform" select widget. There are duplicate choices. E.g., `Windows (x64, 64-bit)` appears twice. I believe this over-counting happens because there are two suffixes for Windows: `tgz` and `exe`. Though I think it makes more sense to show these variants in the "Variant" select widgets (and in fact it already does this).
3. Choose `Windows (x64, 64-bit)`. There are four variants: `Regular`, `Regular`, `CS`, and `CS`. Again, this happens because for each variant, there are two suffixes. It would be nice to display `Regular (tgz)`, `Regular (exe)`, `CS (tgz)`, `CS (exe)` instead.

sorawee:
This PR fixes three bugs:
- Remove an extra "Variant" widget.
- Remove redundant "Platform" options
- Add extension information to variants

Additionally, it replaces the current implementation to a declarative approach
similar to big-bang / React framework. The current approach is _extremely_
complicated due to its imperative nature, generating multiple selection widgets,
embedding several information in non-standard attributes of DOM elements,
and showing/hiding them via JS. This is very error prone due to multiple levels
of the selection widgets, and how the number of levels is dynamic. Additionally,
it is very difficult to modify this page.

The new approach is similar to big-bang. We supply the framework the initial
state and the toDraw function, which constructs the DOM tree based on the input
state. Each DOM element can have a handler (such as onclick) which will
transition the current state into the next state. This makes it much easier to
reason about it.

### DIFF
--- a/download/data.rkt
+++ b/download/data.rkt
@@ -24,7 +24,7 @@
                  [(equal? cpu "ppc") "PPC"]
                  [(equal? cpu "i386") "Intel 32-bit"]
                  [(equal? cpu "x86_64") "Intel 64-bit"]
-                 [else (error "unregonized cpu!")])))]
+                 [else (error "unrecognized cpu!")])))]
     ["(ppc|68k)-mac-classic" "Macintosh Classic (\\1)"]
     ["(ppc|i386)-darwin"
      ,(λ (_ cpu)
--- a/download/download-pages.rkt
+++ b/download/download-pages.rkt
@@ -1,9 +1,10 @@
 #lang plt-web
 
-(require "resources.rkt" "data.rkt" "installer-pages.rkt" "symlinks.rkt"
+(require "resources.rkt" "data.rkt" "installer-pages.rkt" "util.rkt"
+         racket/dict
          plt-web/style
          version/utils
-         net/base64
+         json
          (prefix-in pre: "../minis/pre.rkt"))
 
 (define docs "docs")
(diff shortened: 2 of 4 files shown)


sorawee:
Before:
Screen Shot 2020-06-01 at 14 04 53

After:
Screen Shot 2020-06-01 at 14 05 03

---------------

Before:
Screen Shot 2020-06-01 at 14 04 40

After:
Screen Shot 2020-06-01 at 14 04 30

---------------

Before:
Screen Shot 2020-06-01 at 14 04 19

After:
Screen Shot 2020-06-01 at 14 03 57

sorawee:
Also note that rolled my own big-bang and didn't use React (even though React is really what we should be using) because it doesn't support very old browsers. My big-bang implementation only uses standard JS features which should support IE8 at the very least (and I think it should support IE6 too).

mflatt:
This works ok for me so far, and I can easily believe that it's better code.

Why the "(dmg)" or "(exe)" or "(sh)" in the "Variant" menu? Since that file extension is shown in the link immediately below, it seems redundant.

sorawee:
See the last before-after comparison. While it's true the the extension is clear when it's already chosen, it's not clear when it's about to be chosen. That means users need to guess which one is the right one, and if they guess wrong, they have to try again.

mflatt:
Ah, I see. But the "(dmg)" or "(exe)" is noisy most of the time, when all the choices have the same suffix. Also, "(exe)" vs. "(tgz)" are not different "variants" in the sense of BC vs. CS, so it doesn't seem like the right place for that choice.

The old intent was that the "tgz" option wouldn't show up in the menus, and it was instead left to "More Variants and Checksums" (but "variant" is perhaps not the right word anymore if it's going to be BC vs. CS), and that still seems better to me.

sorawee:
OK, I can fix that tonight. With the new framework, this should be easy to do.

sorawee:
Just to be clear, my plan is to add another level of selection widget named "Extension". If you have a better idea, let me know!

mflatt:
"Extension" is not a good name, because it's not a choice about a file extension. If there's going to be a choice available, the presentation will need more thought.

The choice is between (variously) an installer or disk image and a tarball. "Installer" seems clear enough. "Disk image" is accurate for Mac, but I'm not sure that's terminology that a novice Mac user will recognize, so maybe "installer" is a better approximation. And maybe "archive" is a good word for the second option? Maybe, also, there should be some explanatory text that show up for the "archive" option — something that says it's a directory/folder that can be unpacked and run in place.

But I still think it's better to omit the choice. Especially since there are other distribution options that are not covered by these choices, it's simpler to leave all of the non-obvious ones to "More Varians and Checksums".

mflatt:
What would be a better word than "variant" (which is generic and overloaded) for the "Regular" vs. "CS" choice?  It's a choice of runtime systems, but "Runtime" or "VM" would be too obscure for most users.

sorawee:
I think we should make "Variant" means "Regular" vs "CS". The documentation has been using "Variant" in this sense.

"More Variants and Checksums" then could be changed to "More Installers and Checksums".

mflatt:
Ok, changing to "More Installers and Checksums" makes sense to me.

sorawee:
> Especially since there are other distribution options that are not covered by these choices

I'm not aware of distribution options that are not covered by the current interface. Could you give me some examples?

My initial thought is that because the selection widgets are now much easier to manipulate, showing all installers in the current interface, defaulting to the "obvious ones" is easily doable, and it makes it easier to discover "non-obvious ones". Following this thought, "More Variants/Installers and Checksums" would be changed to simply "Checksums".

mflatt:
I think there's currently no way to get to the "natipkg" installers.

I'll yield to other opinions, but "More Installers and Checksums" seems to me a much more discoverable presentation. Maybe a useful direction is to make "More Installers and Checksums" itself more discoverable.

sorawee:
Done. It turns out that the situation where there could be more than one installers for each variant is when they are `exe` and `tgz`, so for these cases, I added a filter to discard anything that is not `exe` out.

Also renamed the link to "More Installers and Checksums".

mflatt:
This looks good to me, so I'm happy to merge.

Instead of filtering the extra `tgz`s, though, it occurs to me that maybe the problem is that they shouldn't have been listed in `installers.txt` in the first place. @jbclements - any opinion on that?

jbclements:
Thanks for the bump. I filter the installers using a program called filter-installers.rkt, which I see is non-public because it's in the release-pltbuild repo rather than in the racket-lang-org repo. It could easily live there instead. A quick look suggests that I could maybe just drop it in the `download` directory, if that makes sense to you. 

...

Okay, I've done that. Also, I see that there was a bug, in that the file was looking for files containing "windows" rather than "win32". Mea culpa.

PR #124, opened 2020-06-04
Make the download page work on IE7

sorawee:
- List higher-order functions (map, filter, forEach) are not available on IE,
so implement them ourselves.

- Fix a bug in attachEvent. The event name must have the prefix 'on'.

- Use e.srcElement instead of e.target for IE.

- IE hates trailing commas, so remove them.

### DIFF
--- a/download/download-pages.rkt
+++ b/download/download-pages.rkt
@@ -131,6 +131,33 @@
            before installing other packages.}
     @script/inline[type: 'text/javascript]{
 
+// higher-order functions
+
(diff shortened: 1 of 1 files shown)


sorawee:
Fun fact: a trailing comma on an object literal is a syntax error. A trailing comma on an array literal adds `undefined` to the end of the array... Took a while for me to figure that out...

samth:
Maybe we should have a discussion similar to racket/scribble#240 for this as well.

PR #129, opened 2020-08-17
linkify + fix typos

sorawee:


sorawee:
@rfindler: do you know what's going on here? Why are there now 28 commits?

PR #135, opened 2020-08-18
Wip 2nd suggestions

abmclin:
I have made separate commit of each suggested change.

The `#lang typed/racket` example on the front page doesn't run as-is due to a missing parentheses, the fix is provided in one of the commits.

To get the build process working on my machine, I had to replace `rename-file-or-directory` with `copy-directory/files` since the former doesn't work if `/var/tmp` is on a different partition.

Feel free to pick any commits you think are useful.

### DIFF
--- a/www/css/styles.css.pp
+++ b/www/css/styles.css.pp
@@ -505,6 +505,10 @@ a, a:hover {
   transition: color 0.2s;
 }
 
+langwww {
+  color: rgb(115,115,115);
+  font-weight: bold;
+}
 
 .langwww {
     ~height: 11.5rem;
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ To build:
 
 * The simplest way to build the whole site is with 
 
+  [the below command requires you to first install the racket-lang-org directory as a package]
+
   racket -l- racket-lang-org/sync --save-temps --render-locally Web ; open Web/www/index.html
 
   This renders the site in some temp directory and then moves the directory
--- a/sync.rkt
+++ b/sync.rkt
@@ -137,7 +137,7 @@
       (delete-directory/files render-locally?))
     (define tmp-src (build-path tmp-dir "generated"))
     (define local-dest render-locally?)
-    (rename-file-or-directory tmp-src local-dest #t)
+    (copy-directory/files tmp-src local-dest #:preserve-links? #t)
 
     ;; generate and move code into place
     #;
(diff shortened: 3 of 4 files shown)


sorawee:
Shouldn't `tmp-src` be deleted afterward, then?

abmclin:
Not if `--save-temps` is specified on the line? Shouldn't leaving it out take care of the `tmp-src` cleanup?

abmclin:
Looking at the `cond` expression controlling how the command-line options are used, I don't think it works the way it's supposed to. The `(not save-temps?)` and the code moving the rendered files are in separate branches so if `---save-temps` is not given, then `tmp-src` is deleted but the rendered files don't get moved since the `else` branch is not executed.

sorawee:
Yes, this looks more like the intended behavior.

sorawee:
Actually, what should happen when both `save-temps?` and `render-locally?` are `#f`. The old code would delete the temp directory, while the new code would not, right?

abmclin:
yup you're right, let me see...

www/css/styles.css.pp
```diff
@@ -505,6 +505,10 @@
   transition: color 0.2s;
 }
 
+langwww {
+  color: rgb(115,115,115);
+  font-weight: bold;
+}
```
rfindler, line 512: I don't understand CSS that well, but I would have thought that this needs a `.` in front of it?

Also, `langwww` has an opacity of .5 below which is what makes it dark. Maybe it would work better to change that instead?

sorawee: @rfindler actually, this is "correct" in a sense that there's really a tag named `langwww`. So this CSS is indeed correct. But what should really happen is that Pollen shouldn't generate `langwww`. Instead, it should be a `div` with a class `langwww`.

sorawee: To elaborate, the fix is to include 

```
◊(require pollen/tag)
◊(define langwww (default-tag-function 'div #:class "langwww"))
```

in either `www/index.html.pm` or `www/util.rkt`.

rfindler: Thanks!

abmclin: What sorawee said. I wasn't sure if the `langwww` tag was really intended since it's not a standard HTML tag but technically it's possible to extend HTML with additional tags...

There is a `.langwww` CSS rule present in styles.css.pp which applies to `langwww` class and it's not being used at all.

abmclin:
I have just pushed more changes which I believe should give the correct behavior for various combinations of `--save-temps` and `--render-locally` but I can't test all of them since I don't have a `.aws-keys` file on my system.

www/css/styles.css.pp
rfindler: Okay, I didn't like the boldface everywhere so I went with a darker color. [What do you think?](http://www.cs.northwestern.edu/~robby/tmp/Web/www)

abmclin: It looks better already, but would you mind changing the rgb values to `rgb(100,100,100)` pretty please? :-)

rfindler: Done!

PR #150, opened 2020-11-01
Add Chinese to translations

tsao-chi:


jessealama:
@jeapostrophe LGTM

rfindler:
Is this the best translation to use? I wonder if 世界好 or even 大家好 might be more accurate?

I'm totally fine with the current change, just wanted to offer a suggestion.

PR #152, opened 2021-01-15
download: fix <select> not reset on Firefox

sorawee:
On Firefox, setAttribute('selected') can't be used to programmatically
update the page. Apparently, this appears to be the correct behavior
according to the HTML specification, and other browsers (e.g., Chrome)
do it wrong.
(See https://github.com/jsdom/jsdom/issues/1130#issuecomment-102375898)
In any case, the solution here is to mutate the HTML object directly
(that is, change the property, not the attribute).

CC: @mflatt

### DIFF
--- a/download/download-pages.rkt
+++ b/download/download-pages.rkt
@@ -179,6 +179,8 @@ var bigbang = null;
 
 var elem = null;
 
+var property = null;
+
 (function() {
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #158, opened 2021-05-23
Use Racket on macOS Big Sur 11.1 or above

shhyou:
It's easier to point the students using Mac M1 to a fixed stable version such as 8.1 or 7.9. For this reason, mentioning the snapshot builds creates extra confusion when the particular build happens to contain other bugs.

As Racket works on Mac M1 with macOS 11.1 (or above) but not macOS 11.0, explain that the user may need to upgrade their OS.

### DIFF
--- a/download/download-pages.rkt
+++ b/download/download-pages.rkt
@@ -114,11 +114,13 @@
                       @pre:installers{snapshot build}.}}
            null)
       @(cond
-         [(equal? version version-before-m1-support)
+         [(version<=? version version-before-m1-support)
           @div[id: "m1_mac_explain"
                style: note-style]{
                @div{@b{M1 Mac users:}}
-               @list{Try a 64-bit ARM build from one of the @pre:installers{snapshot sites}.}}]
+               @list{The latest version of Racket supports Apple Silicon directly.
+
+               For version @version, the Mac OS (Intel 64-bit) variant requires macOS Big Sur 11.1 or above.}}]
          [(version<? version-before-m1-support version)
           @div[id: "m1_mac_explain"
                style: note-style]{


download/download-pages.rkt
```diff
@@ -118,7 +118,12 @@
           @div[id: "m1_mac_explain"
                style: note-style]{
                @div{@b{M1 Mac users:}}
-               @list{Try a 64-bit ARM build from one of the @pre:installers{snapshot sites}.}}]
+               @list{
+               Try upgrade to macOS Big Sur 11.1 or above and select
```
rfindler, line 122: typo: "upgrading"

shhyou:
Thanks! I also rearranged the sentence a little bit.

rfindler:
Here's a variant that tries to be softer about forcing them to upgrade their operating system.

```
The Mac OS (Intel 64-bit) variant works on macOS Big Sur 11.1 or above. 
If it crashes, consider upgrading your operating system.

Alternatively, newer versions of Racket support Apple Silicon directly.
```

Also, why are we pointing to the snapshot sites? I'm wondering if that's old text so I just changed it.

mflatt:
I agree that pointing to the snapshots is obsolete, and the suggested text to point to a later Racket version seems right.

shhyou:
> Also, why are we pointing to the snapshot sites? I'm wondering if that's old text so I just changed it.

I think (but didn't check) when that paragraph was added, macOS 11.1 hasn't been released yet.

rfindler:
LGTM

download/download-pages.rkt
```diff
@@ -118,7 +118,10 @@
           @div[id: "m1_mac_explain"
                style: note-style]{
                @div{@b{M1 Mac users:}}
-               @list{Try a 64-bit ARM build from one of the @pre:installers{snapshot sites}.}}]
+               @list{The Mac OS (Intel 64-bit) variant works on macOS Big Sur 11.1 or above.
+               If it crashes, consider upgrading your operating system.
+
+               Alternatively, newer versions of Racket support Apple Silicon directly.}}]
```
samth, line 124: Could we say "current" here? The problem is people trying to download 8.0, right?

shhyou:
Done! ("newer" -> "current")

I don't have opinion between "newer versions" and "the current version". But I don't understand the comment on 8.0. Are you thinking about HtDP?

samth:
I'm trying to understand who this is for. Is this for someone going to download.racket-lang.org with an 11.0 M1 machine? Or for people who have been linked directly to a past download page? Or something else I'm failing to understand?

sorawee:
It's for a past download page. In particular, it appears on https://download.racket-lang.org/racket-v7.9.html, and only for version 7.9 (that is, e.g., https://download.racket-lang.org/racket-v7.8.html doesn't have this notice).

I think `(equal? version version-before-m1-support)` should be changed to `(version<=? version version-before-m1-support)` (or something like that) so that the warning about Big Sur 11.1 will appear to older versions as well.

I don't like the word "current version", because when users visit a page like https://download.racket-lang.org/racket-v7.9.html, the current version could mean the version the user is visiting. What about saying "8.0 or later"? Or even "latest version".

rfindler:
> But I don't understand the comment on 8.0. Are you thinking about HtDP?

I'm not sure if this is directed at me, but if it is: I don't know! :) I'm happy with the changes.

shhyou:
> I'm trying to understand who this is for. Is this for someone going to download.racket-lang.org with an 11.0 M1 machine? Or for people who have been linked directly to a past download page? Or something else I'm failing to understand?

For M1 users with macOS 11.0 using Racekt 7.9. In that case, the Intel 64-bit installation crashed, and the snapshot build gave [this error](https://github.com/racket/htdp/issues/143).

samth:
If this is a warning only for the 7.9 page, then we should more strongly encourage downloading the latest version, I think.

shhyou:
How about this?

- Point to the latest version first
- Hints that the Intel variant needs Big Sur 11.1 or above

PR #161, opened 2021-06-20
lists: use google group names when constructing subscribe addrs

Bogdanp:
Re. https://github.com/racket/racket-lang-org/commit/db9957dcfd7d3b06b8b1357a1e3ad066177a7245#r52404666


Bogdanp:
👋 ping

PR #163, opened 2021-11-07
Add missing entries to Vibrant Community tab

spdegabrielle:


spdegabrielle:
@jbclements

spdegabrielle:
Need to address the two community sections,
But discourse was added to the second section only so I’ll redo and resubmit in a new pr

PR #172, opened 2022-01-12
"Jet Fueled" -> "Practical"

mflatt:
The "Jet Fueled" label doesn't make sense to me, and as a label it seems unlike others (especially if #171 is accepted). I think the panel is about how Racket is "Practical", so I'm proposing that change.


mfelleisen (commented):
Well ...

rfindler:
🚀

PR #180, opened 2022-06-30
Add simple "build everything" GitHub Action

jessealama:
As a simple sanity check, we should be able to build everything (all pages on all sites). This GitHub Action will help us track whether we've busted a site.


jessealama:
FYI this is just a QA check. Nothing gets uploaded or changed anywhere. We just build everything in a fresh directory, as if one were working privately on one's own laptop.

mflatt:
This makes sense to me.

PR #184, opened 2022-07-07
Tidy CSS, fix quote blooper, simplify speaker setup

jessealama:


jessealama:
@sorawee This contains the fix for the quote problem you found.

PR #187, opened 2022-07-13
Add Cameron Moy

jessealama:
In addition, unbreak the "does it all build?" GitHub Action. (Interestingly, [a post on racket-users](https://groups.google.com/g/racket-users/c/veeoUnsj-ow) pointed the way to a solution.)


jessealama:
@mflatt Could you deploy this change?

PR #192, opened 2022-07-23
Add title & abstract for Marco Morazán

jessealama:
Also, start to flesh out some metadata for the event.


jessealama:
@mflatt Can you deploy this when you have a moment?

PR #207, opened 2022-09-19
rcon: Add an abstract and some bios

jessealama:


jessealama:
@mflatt Could you deploy when you have a moment?

PR #212, opened 2022-10-06
rcon: Set up initial schedule for talks, add ticket purchase link, and update hotel info

jessealama:


jessealama:
@mflatt Could you please deploy this when you have a moment? We finally have a link where people can purchase tickets. (So far, people could only express interest in attending. We haven't had a complete registration process that included purchasing a ticket. But now people can actually get a ticket.)

PR #214, opened 2022-10-16
rcon 2022: Add microformat semantic data for speakers

jessealama:


jessealama:
@mflatt Would you mind deploying this when you have a moment?

PR #215, opened 2022-10-16
rcon: Add speaker affiliation as microformat data

jessealama:


jessealama:
Abandoning because I screwed up git. Will re-open a clean PR later.

PR #216, opened 2022-10-16
rcon: Add affiliation data as microformat

jessealama:


jessealama:
@mflatt Could you deploy this when you have a moment?

PR #219, opened 2022-10-20
rcon: Fix schedule

jessealama:


jessealama:
@mflatt Would you mind deploying this? I've tweaked the schedule a bit and updated the description of the hackathon.

Issue #228, opened 2022-11-23
Add Scribble styles to blog pages


jryans:
I was reading the recent blog post on [Shallow and Optional Typed Racket](https://blog.racket-lang.org/2022/11/shallow-typed-racket.html), and I was a bit surprised to see the code examples rendered in the same style as the prose:

<img width="669" alt="image" src="https://user-images.githubusercontent.com/279572/203662065-27ee55eb-2a71-4a79-b05b-0be48805b008.png">

It seems especially unfortunate since the blog post is already written in Scribble and attempting to mark those blocks as code, but the blog doesn't seem to have any corresponding CSS.

It seems like pulling in some amount of Scribble styles onto the blog would make this look much nicer... should I give that a try, or is there some reason to leave things as they are?

bennn:
Go for it!

This seems like a bug to me. When I wrote the post and built locally (`cd blog; raco frog -bp`), the code blocks were styled as code:

![Screen Shot 2022-11-28 at 21 12 07](https://user-images.githubusercontent.com/1731829/204421855-6f758019-ee1a-4d13-97e5-0fb43880a8b3.png)

jryans:
@bennn, thanks for taking a look. It seems like your local `main.css` was generated just before commit https://github.com/racket/racket-lang-org/commit/51eb3e0f5d6bda32740ea0c7e1422a373a7e89b7 which reverted some Scribble-related styles. I guess @LeifAndersen committed them accidentally, but it's hard to be sure, and also it was quite a while ago...

I'll prepare a PR that extracts the key bits of styling used on other Scribble pages without interfering with the overall look and feel of the blog.

PR #231, opened 2022-11-29
Style Racket code blocks used on the blog

jryans:
This copies in the essential styles from Scribble to present Racket code blocks with nice formatting on the blog. In particular, this ensures they are shown with monospaced fonts and use the font colours from the documentation.

The styles added here are deliberately kept minimal to avoid changing the blog's overall look and feel outside of Racket code blocks.

image

Fixes https://github.com/racket/racket-lang-org/issues/228

cc @samth @bennn


mflatt:
Thanks!

PR #232, opened 2022-12-04
Announce Racketfest 2023

jessealama:


jessealama:
@mflatt What do you think? I don't want to unilaterally merge this PR if there's consensus that this isn't OK.

mflatt:
I'm in favor!

PR #234, opened 2022-12-12
update FEP to include unacceptable behaviours.

spdegabrielle:
Making the list of unacceptable behaviours a reference to another document just makes it harder to read and adds a dependency on the acm website.  This PR copies the current text at https://www.acm.org/about-acm/policy-against-harassment verbatim.

### DIFF
--- a/www/friendly.html.pm
+++ b/www/friendly.html.pm
@@ -11,7 +11,63 @@ The Racket community aims to improve the world through programming. It started w
 
 Racket project management wants to encourage and preserve this open exchange of ideas, which requires an environment that enables all to participate without fear of personal harassment. We define ◊em{harassment} to include specific unacceptable factors and behaviors listed in ◊link["https://www.acm.org/about-acm/policy-against-harassment"]{the ACM's policy against harassment}. Unacceptable behavior will not be tolerated.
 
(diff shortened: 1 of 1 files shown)


joelmccracken:
Are the bidirectional unicode characters embeded in this PR intentional? See the warning under "Files changed"; looks like a weird copy/paste thing from the acm site

mfelleisen:
@spdegabrielle {{ Sorry I am buried in work this week.  I didn't notice your request until now -- by accident . ]]

I agree with you that an exploit list is overburdening the statement. We are friendly _community_.  (Yes, I have been accosted on occasion on issues that are _not_ listed. I have a thick skin, so I would never ever report something like that.  if I had a thin skin, the list might deter me from reporting it.)

spdegabrielle:
@mfelleisen  Thanks Matthias, no need to apologise.


> on issues that are not listed

this is indicates to me the reference to the ACM list is not sufficient, and further reason to update this to meet the needs of everyone. 

My goal is to make it clear that everyone is welcome, and set clear boundaries, so any undesirable behaviours are addressed quickly, before they can escalate or damage the parties involved.

Before we commit to this PR, would it be better to look at the Rust, Go and Julia* to see if they are a better fit?

If you think that is a good approach I'll raise an issue with those lists as a starting point, before jumping into a PR. 

Please let me know how you want to move forward.

best regards

Stephen

[* I deliberately choose other open source *language* communities that I believe are well regarded]

PS 
Thanks @joelmccracken for spotting the *bidirectional unicode characters*. Another reason not to depend on the ACM website.

PR #238, opened 2023-02-22
Add community moderation team to FEP

spdegabrielle:
Update to add community moderation team

### DIFF
--- a/www/friendly.html.pm
+++ b/www/friendly.html.pm
@@ -1,6 +1,10 @@
 #lang pollen
 
-◊(require (only-in "index.html.pm" top))
+◊(require (only-in "index.html.pm" top)
+          racket/string)
+
+◊(define (email user . host)
+   (format "~a at ~a" user (string-join host " dot ")))
 
 ◊(top)
 
@@ -13,18 +17,14 @@ Racket project management wants to encourage and preserve this open exchange of 
 
 Anyone witnessing or subject to unacceptable behaviour should notify the moderation team or Racket Project management.
 
-Community moderation team:
-
-◊link["https://github.com/countvajhula"]{Siddhartha Kasivajhula}
-
-◊link["https://github.com/jackfirth"]{Jack Firth}
-
-◊link["https://github.com/samdphillips"]{Sam Phillips}
-
-◊link["https://github.com/spdegabrielle"]{Stephen De Gabrielle} (◊link["mailto:spdegabrielle@gmail.com"]{spdegabrielle@gmail.com})
-
-◊link["https://github.com/jryans"]{J. Ryan Stinnett} (◊link["mailto:jryans@gmail.com"]{jryans@gmail.com})
-
+◊div{Community moderation team:}
+◊div[#:style "padding:0.5rem"]{
+◊link["https://github.com/countvajhula"]{Siddhartha Kasivajhula} (◊email["sid" "countvajhula" "com"])
+ ◊link["https://github.com/jackfirth"]{Jack Firth} (◊email["jackhfirth" "gmail" "com"])
+ ◊link["https://github.com/samdphillips"]{Sam Phillips} (◊email["samdphillips" "gmail" "com"])
+ ◊link["https://github.com/spdegabrielle"]{Stephen De Gabrielle} (◊email["spdegabrielle" "gmail" "com"])
+ ◊link["https://github.com/jryans"]{J. Ryan Stinnett} (◊email["jryans" "gmail" "com"])
+}
 
 ◊link["team.html#management"]{Racket project management} can be reached at ◊link["mailto:management@racket-lang.org"]{management@racket-lang.org}.
 
--- a/www/friendly.html.pm
+++ b/www/friendly.html.pm
@@ -23,7 +23,7 @@ Community moderation team:
 
 ◊link["https://github.com/spdegabrielle"]{Stephen De Gabrielle}
 
-◊link["https://github.com/jryans"]{J. Ryan Stinnett}
+◊link["https://github.com/jryans"]{J. Ryan Stinnett} (◊link["mailto:jryans@gmail.com"]{jryans@gmail.com})
 
 
 ◊link["team.html#management"]{Racket project management} can be reached at ◊link["mailto:management@racket-lang.org"]{management@racket-lang.org}.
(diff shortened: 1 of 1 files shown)


spdegabrielle:
Draft PR. 
Before requesting review I'll ask @samdphillips, @jackfirth & @countvajhula to confirm they are happy with this.

samdphillips:
👍🏼

sorawee:
The paragraphs above and below "Community moderation team" are about Racket project management, so I don't think this is a right place to add it. It could mislead people to think that "Community moderation team" is a part of the Racket project management (which is not the case in my understanding).

What exactly are you trying to accomplish by providing this list? If it's so that people can contact the team, I don't think linking to GitHub is a good idea. There's no messaging functionality in GitHub.

jackfirth:
> The paragraphs above and below "Community moderation team" are about Racket project management, so I don't think this is a right place to add it. It could mislead people to think that "Community moderation team" is a part of the Racket project management (which is not the case in my understanding).
> 
> What exactly are you trying to accomplish by providing this list? If it's so that people can contact the team, I don't think linking to GitHub is a good idea. There's no messaging functionality in GitHub. 

GitHub profiles can optionally show email addresses

countvajhula:
Looks like Jack and I display email addresses while Stephen has it in his bio, while Sam doesn't show an email. I agree with @sorawee though that linking to Github page doesn't immediately suggest that it is a way to contact the person but rather just a link to their developer profile. I would favor something like name at email dot com style obfuscated emails. Although the emails sure look like they're non-obfuscated in our Github profiles so I don't know if it's even worth bothering with obfuscation nowadays. Is this not a thing anymore?

www/friendly.html.pm
```diff
@@ -11,7 +11,20 @@ The Racket community aims to improve the world through programming. It started w
...
+◊link["https://github.com/countvajhula"]{Siddhartha Kasivajhula}
+
+◊link["https://github.com/jackfirth"]{Jack Firth}
+
+◊link["https://github.com/samdphillips"]{Sam Phillips}
+
+◊link["https://github.com/spdegabrielle"]{Stephen De Gabrielle}
+
+
+◊link["team.html#management"]{Racket project management} can be reached at ◊link["mailto:management@racket-lang.org"]{management@racket-lang.org}.
```
countvajhula, line 27: Re: possible confusion with project management and community moderators, how about retaining the references to "project management" in these changes, but rewording the paragraph above to something like:

_In order to encourage and preserve this open exchange of ideas, we seek to foster an environment that enables all to participate without fear of personal harassment. We define ◊em{harassment} to include specific unacceptable factors and behaviors listed in ◊link["https://www.acm.org/about-acm/policy-against-harassment"]{the ACM's policy against harassment}. Unacceptable behavior will not be tolerated._

That way the only references to project management make it clear that they are another recourse and not the subject here.

spdegabrielle: I like the suggested text, but I belive the risk of confusion is low.

Can we review this textual change as a follow-up PR, once this one is approved?

**Racket project management wants** 
_to encourage and preserve this open exchange of ideas,_ 
**which requires an** 
_environment that enables all to participate without fear of personal harassment._

**In order** 
_to encourage and preserve this open exchange of ideas,_ 
**we seek to foster an** 
_environment that enables all to participate without fear of personal harassment._

jryans:
(I was just added to the moderation team, thanks all! 🙂) 

Displaying emails directly does seem a bit more actionable for anyone making a report. Personally, I would be fine with my own email address being displayed next to my name (with or without obfuscation, whatever the group prefers).

spdegabrielle:
> What exactly are you trying to accomplish by providing this list?

I approached the Racket leadership team proposing that I ask the current discord moderators if they are willing to be listed as community moderators, and listed explicitly as part of the friendly environment policy. 

I believe 
* this is an important step for the Racket community,
* this will help build confidence that this an OSS PL community that cares about all participants, 
* and this demonstrates community is willing to change when needed.

Other OSS PL communities also have community moderators.
http://julialang.org/community/standards/
https://www.rust-lang.org/policies/code-of-conduct

spdegabrielle:
I believe this is now comepete. I’ll pass to the team for review
@jbclements @jeapostrophe @mfelleisen @mflatt @rfindler @samth

www/friendly.html.pm
rfindler: I think that obfuscating the email addresses in this is no barrier to the spammers, only a (mild) barrier to the legit folks who want to contact the moderation team. If it were me, I'd put in `mailto:` links.

www/friendly.html.pm
```diff
@@ -1,6 +1,10 @@
 #lang pollen
 
-◊(require (only-in "index.html.pm" top))
+◊(require (only-in "index.html.pm" top)
+          racket/string)
+
+◊(define (email user . host)
+   (format "~a at ~a" user (string-join host " dot ")))
```
rfindler, line 8: I think that obfuscating the email addresses in this is no barrier to the spammers, only a (mild) barrier to the legit folks who want to contact the moderation team. If it were me, I'd put in mailto: links.

mflatt: From the discussion above, though, at least two of the owners of the email addresses wanted light obfuscation.

rfindler: I saw one person saying both they'd do it that way on their own and also questioning whether it was a good idea to do that way; I didn't see a second person. (I must have missed a comment.) Anyway, I'm not one of the people whose emails are here, tho. I just posted because it seems like making it easy to contact these folks is worth doing.

spdegabrielle: I’m not bothered. My email is easily found.

mflatt: After checking with Sam, it sounds like everyone is happy with mailto links after all, so I pushed that change.

rfindler: Thanks!

countvajhula: Works for me!

countvajhula (approved):

www/friendly.html.pm
spdegabrielle: @countvajhula created a new issue so this feedback isn't lost and can be addressed later #239

spdegabrielle:
Please forgive my impatience - are there any outstanding issues with this PR?

mflatt:
It seems that everyone is happy, so I'll merge.

mfelleisen:
Well, "happy" is one thing, "quiet" is another :-)

Issue #240, opened 2023-03-17
racketcon 2022 link broken 404


jestarray:
https://con.racket-lang.org/2022 Visiting this url gives me a 404 uncaught exception

spdegabrielle:
Folder looks ok
https://github.com/racket/racket-lang-org/blob/master/rcon/2022/index.rkt

I’m not clear how https://con.racket-lang.org/2022 is configured and built?

sorawee:
That appears to be intentional. 

https://github.com/racket/racket-lang-org/blob/65c54d3770f332ee3e658bc515b219643d37ffb1/annual-utils.rkt#L39

For the current year, `index.html` won't be copied over. Instead, you can access it via the "main" site: https://con.racket-lang.org/

spdegabrielle:
Lol
2022 at https://con.racket-lang.org/

sorawee:
I do think that it's supposed to work though.

The problem appears to be due to https://github.com/racket/racket-lang-org/blob/master/sync.rkt#L132, which should update AWS configuration to redirect correctly. @jessealama updated the script for 2022, but I believe the script wasn't actually run. As a result, the config on AWS is still from 2021. And we can see this:

- https://con.racket-lang.org/2021/index.html is redirected to https://con.racket-lang.org/.
- https://con.racket-lang.org/2021/ works.

PR #245, opened 2023-07-03
rcon: Let's roll, 2023

jessealama:
Add a couple of talks (more to come soon). This is just preliminary skeletal information that will get fleshed out later.

### DIFF
--- a/rcon/2023/.gitignore
+++ b/rcon/2023/.gitignore
@@ -1,0 +1,1 @@
+/index.html
--- a/rcon/2023/index.rkt
+++ b/rcon/2023/index.rkt
@@ -1,0 +1,347 @@
+#lang at-exp racket/base
+
+(require racket/match
+         racket/string
+         xml
+         txexpr
+         (prefix-in gregor: gregor)
+         "lib.rkt")
+
+(define-div column
+  ,@centered
+  [width "45em"])
+
+(define-div banner
+  ,@centered)
+
+(define-div title-append
+  [display inline-block]
+  (margin-left auto)
+  (margin-right auto)
+  [text-align left]
+  [margin-bottom "1em"])
+
+(define-div pagetitle
+  [font-size "64pt"]
+  [font-family "'Source Code Pro', monospace"])
+
+(define-div main
+  [font-family "'Raleway', sans-serif"])
+
+(define header-font
+  `([font-weight "bold"]))
+
+(define-div subtitle
+  ,@centered
+  [font-size "40pt"]
+  ,@header-font)
+
+(define-div subsubtitle
+  ,@centered
+  [font-size "20pt"]
+  ,@header-font)
+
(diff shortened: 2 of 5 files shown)


jessealama:
@mflatt How does this look?

jessealama:
@rfindler FYI

mflatt:
Ok!

As a minor point, I think it's better to not use `#:current #t` and rely on a copy that is created further below in `all.rkt` to populate `index.html`. That way, the page is at both `index.html` and (as a kind of permalink) at `2023/index.html`. If that small change sounds ok and this is otherwise ready to post, I can make that adjustment while also updating the main Racket page to link.

mflatt:
It might also be better to adjust the style in some way, such as font or color. It's easy to lose track of which year you're looking at when they all look the same. (I almost suggested a correction to the date, then realized that I was looking at the wrong year.)

Do you have opinions on what the style should be, or should I try things to make a suggestion?

mflatt:
It's easy to fall into into playing with the page style, and I did... So, here's one suggestion, but feel free to ignore: https://github.com/mflatt/racket-lang-org/tree/2023styleidea

sorawee:
I did a bit of investigation on the "broken links" problem for 2022. It would be nice to make sure this doesn't happen with 2023.

From https://github.com/racket/racket-lang-org/issues/240#issuecomment-1522697457:

> The problem appears to be due to https://github.com/racket/racket-lang-org/blob/master/sync.rkt#L132, which should update AWS configuration to redirect correctly. @jessealama updated the script for 2022, but I believe the script wasn't actually run. As a result, the config on AWS is still from 2021. And we can see this:

>    https://con.racket-lang.org/2021/index.html is redirected to https://con.racket-lang.org/.
>    https://con.racket-lang.org/2021/ works.

mflatt:
> I did a bit of investigation on the "broken links" problem for 2022. It would be nice to make sure this doesn't happen with 2023.

There's a potential repair hiding in my style-suggestion branch: https://github.com/mflatt/racket-lang-org/commit/13efb3b3bb5189d686796a8fff0f15973800673a#diff-b3bc884c5249a4b316c7c8960240b4426fbd21238ba6a5207c54d7755bfdf010R67. I'll make sure to copy that over to whatever ends up being the new page.

jessealama:
Looks great; thanks for the style help! Let's roll with your change.

jessealama:
> As a minor point, I think it's better to not use `#:current #t` and rely on a copy that is created further below in `all.rkt` to populate `index.html`. That way, the page is at both `index.html` and (as a kind of permalink) at `2023/index.html`. If that small change sounds ok and this is otherwise ready to post, I can make that adjustment while also updating the main Racket page to link.

I'm in favor of that. I agree there's no need to make `index.html` and `2023/index.html` the same.

PR #248, opened 2023-07-15
rcon 2023: Add a few more talks

jessealama:
Details coming, but the speakers are confirmed.


jessealama:
@mflatt Could you please deploy this?

PR #250, opened 2023-07-17
rcon: Add a title & abstract, as well as hotel information

jessealama:


jessealama:
@rfindler Do you think we should list the Hilton Orrington as a recommended option before we've reserved a block of rooms there? In my view, that makes sense, though I see the point that we may want to postpone our recommendation until we know that a block of rooms has been reserved for us. (The counterargument is that we can't generally guarantee that someone will get the reserved rate, since the hotel may fill up or RacketCon 2023 may be unexpectedly popular.)

rfindler:
Let's see what Wynante thinks when we talk tonight?

Robby

PR #251, opened 2023-07-21
feat: support python3

sorawee:
This commit adds a support for Python 3 by looking for `python3` in case the `python` executable can't be found.

### DIFF
--- a/blog/all.rkt
+++ b/blog/all.rkt
@@ -4,6 +4,7 @@
          racket/path
          racket/system
          raco/all-tools
+         "python.rkt"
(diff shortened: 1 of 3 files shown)


rfindler:
Looking at the diff, this looks like a clear improvement to me (and I recently created a symbolic link on my machine for exactly this problem). I didn't try to rebuild the site with the PR, tho. If you've done that @sorawee I'm happy to merge unless there is some other testing we should do?

sorawee:
Yes, I built the site locally and it appears to work as intended.

PR #259, opened 2023-08-25
rcon: Add new abstract

jessealama:


jessealama:
@mflatt could you please deploy this?

PR #273, opened 2023-09-24
"Die Macht der Abstraktion" -> "Schreibe Dein Programm!"

mikesperber:
The new book is effectively a rewrite, and finally done.


mfelleisen:
@mikesperber Mike, don't you want to keep the old book and just add the new one?

mikesperber:
> @mikesperber Mike, don't you want to keep the old book and just add the new one?

@mfelleisen No, just ditch the old one. It's officially deprecated now.

PR #277, opened 2023-10-09
rcon: Fix schedule

jessealama:


jessealama:
@rfindler This schedule should be fixed now.

PR #278, opened 2023-10-09
adjust RacketCon'23 style to make keynote more prominent

mflatt:
Also, add more whitespace between the events. I'm less sure of this change, but it makes things easier to read to me.


mflatt:
@jessealama and @rfindler

mflatt:
Added a few more style adjustments, which includes reducing the added space.

jessealama:
Thanks! Looks good.

PR #281, opened 2023-10-13
RacketCon'23: a livestream and chat iframes

rfindler:
I'm not sure if this is the best way to put the livestream. Right now it is at the top of the page, looking something like this:

Screenshot 2023-10-12 at 7 46 46 PM


rfindler:
Or, maybe just the livestream for now and bring in the chat right before racketcon actually happens?

mflatt:
Yes, waiting on chat makes sense.

I think it also makes sense to wait until closer to the day to add the linvestream display, so the schedule isn't pushed down just yet.

rfindler:
That works!

I've pushed a second commit to disable the chat and we can push the green button here when we're ready to make the livestream appear, then.

jessealama:
+1 to the idea of waiting till the day-of to deploy the chat feature

Otherwise, LGTM, thanks!

jessealama:
With the chat disabled (for now), shall we can deploy this? If so, I suppose we'd need a second pull request (which we'll wait to deploy) where the chat is enabled. Or do we want to wait to merge this until the day-of?

jessealama:
@mflatt Could you deploy this, please?

mflatt:
Ok, done.

PR #283, opened 2023-10-18
rcon: Add another speaker bio

jessealama:


jessealama:
@mflatt Would you mind deploying this change (speaker bios)?

PR #290, opened 2023-11-16
Update languages page

countvajhula:
@spdegabrielle pointed out that this page was out of date and suggested these additions.

This page is still unlinked from the homepage.

### DIFF
--- a/www/languages.html.pm
+++ b/www/languages.html.pm
@@ -18,6 +18,10 @@ The following is a selection of of languages built with Racket including languag
 ◊section{
 ◊div{Languages}
 
+◊link["https://joe.cz/rrll/RRLL__Actor_Basic.html"]{Actor Basic}
+
+A single-value flow-oriented language differentiating between flow steps and value expressions.
+
 ◊link["https://docs.racket-lang.org/algol60/"]{Algol 60}
 
 The “Algol 60” language for DrRacket implements the language defined by the “Revised Report on the Algorithmic Language Algol 60,” edited by Peter Naur.
@@ -29,6 +33,11 @@ Asi64 is a cross-platform 6502 assembler. Primarily aimed at programming the Com
 ◊link["https://docs.racket-lang.org/datalog/"]{Datalog: Deductive Database Programming}
 
 Datalog is a declarative logic language in which each formula is a function-free Horn clause, and every variable in the head of a clause must appear in the body of the clause. Datalog is also a lightweight deductive database system where queries and database updates are expressed in the logic language.
+
+◊link["https://docs.racket-lang.org/forth/"]{Forth}
+
+Forth is a stack-based calculator language. This package implements a subset of Forth as a Racket #lang.
+
 
 ◊link["https://lexi-lambda.github.io/hackett/"]{The Hackett Programming Language}
 
@@ -65,6 +74,11 @@ The Pollen language is markup-based, so you can write & edit text naturally. But
 ProfessorJ is a plug-in for DrRacket that implements variants of Java, especially for teaching purposes.
 
 
+ ◊link["https://docs.racket-lang.org/qi/"]{Qi: An Embeddable Flow-oriented Language}
+
+Qi is a general-purpose language that's useful for programming in the functional, immutable style and easily embedding such computations anywhere in your programs. It embodies a point-free, "flow-oriented" way of thinking that is appropriate for many common classes of problems, including data pipelines and working with multiple values.
+
+
  ◊link["https://github.com/racketscript/racketscript"]{RacketScript}
 
 A lightweight Racket to JavaScript compiler.
@@ -73,6 +87,11 @@ A lightweight Racket to JavaScript compiler.
  ◊link["https://docs.racket-lang.org/rash/"]{Rash: The Reckless Racket Shell}
 
 Rash is a shell language embedded in Racket. It has a concrete syntax that is amenable to quick and easy interactions without lots of punctuation overhead. It aims to allow shell-style interaction and programming to be freely mixed with more general-purpose Racket code.
+
+
+ ◊link["https://docs.racket-lang.org/rhombus/"]{Rhombus}
+
+Rhombus is an experimental, general-purpose programming language with conventional expression syntax that is built on Racket and that is macro-extensible in the same way as Racket.
 
 
  ◊link["https://docs.racket-lang.org/riposte/"]{Riposte}
--- a/www/languages.html.pm
+++ b/www/languages.html.pm
@@ -65,7 +65,7 @@ The Pollen language is markup-based, so you can write & edit text naturally. But
 ProfessorJ is a plug-in for DrRacket that implements variants of Java, especially for teaching purposes.
 
 
- ◊link["https://github.com/vishesh/racketscript"]{Racketscript}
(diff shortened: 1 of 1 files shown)


sorawee:
Please ask @mfelleisen to review this. I don't know if "unlinked from the homepage" is intentional or not, or if these entries are qualified, or if there are any other missing langs (Esterel?).

rfindler:
Esterel is okay not to be included (yet 😄)

mfelleisen:
@sorawee Other than Rhombus I wold not include any of these languages in our home page.

sorawee:
@mfelleisen note that this is not the homepage per se. It's a subpage that is currently orphaned: https://racket-lang.org/languages.html.

So there are two issues to think about.
- Should we even have this subpage? If so, we should make it useful by linking it. Then the question becomes: where should we do that? But if not, we can simply delete it.
- If we will maintain this page, what languages should be included?

mfelleisen:
(When I use the word ‘homepage’ I speak like the old man I am — and mean the whole site.) 

The page was meant to show the variety of DSLs that people can write in Racket. 
It was not meant to be “look ma, people made all these languages in Racket.” 

So my preference would be to ‘freeze’ this page, with the side condition of adding 
any language that it is truly exceptional in some ways. (Of course, this almost means
removing some things that have snuck in.)

countvajhula:
I would empower community members like Stephen (with help from others, as in this case) to make such changes without needing to go through this kind of rigid approval process. It would be great for the website to be a dynamic portal serving the changing needs of an active community, and Stephen is very good at his job.

"The bamboo that bends is stronger than the oak that resists."

spdegabrielle:
@countvajhula you have clearly never had to work with me (insert laughing face emoji here)

I'll address the note from @mfelleisen because he has the historical context that I don't 

> The page was meant to show the variety of DSLs that people can write in Racket.
> It was not meant to be “look ma, people made all these languages in Racket.”
> So my preference would be to ‘freeze’ this page, with the side condition of adding
> any language that it is truly exceptional in some ways. (Of course, this almost means
> removing some things that have snuck in.)

I think this should be a page to showcase interesting languages with as much variety as possible. I dont have a clear definition of exceptional.

https://docs.racket-lang.org/search/index.html?q=H%3A currently returns docs for 237 #lang's  but the presentation is an alphabetical list of (sometime poorly chosen) names with not description - that query by it's nature fails to include languages that are not implemented with the `#lang` mechanism. 

[Algol 60](https://docs.racket-lang.org/algol60/)
- If there is a Racket implementation of a language that has featured in HOPL it deserves to be included. - I think that is reason enough to add the Forth implementation

[Asi64](https://docs.racket-lang.org/asi64/)
An example of a Racket powered assembler targeting an microcontroller. I believe it is unique.

[Datalog: Deductive Database Programming](https://docs.racket-lang.org/datalog/) - I think I saw a paper on datalog at oopsla. Is datalog embedded in clojure?

[The Hackett Programming Language](https://lexi-lambda.github.io/hackett/) - Haskell in Racket!

[The Heresy Programming Language](https://docs.racket-lang.org/heresy/) - Basic in lisp! 

[#lang lindenmayer, a language for L-Systems](https://docs.racket-lang.org/lindenmayer/) - a staple of computer art, recreational computing, it's great and it never goes away.

[Parenlog](https://docs.racket-lang.org/parenlog/) - should be listed as Prolog in a historical important section.


[Pie](https://docs.racket-lang.org/pie/). - notable book

[Pollen](https://docs.racket-lang.org/pollen/). - scribble too surely?

[ProfessorJ: Java in Racket](https://docs.racket-lang.org/profj/) - fits two categories - another implementation of incremental languages for teaching - but also an implementation of a historically important language alongside 

[Racketscript](https://github.com/vishesh/racketscript) - undeniably important and relevant. but the link is wrong it should be https://github.com/racketscript/racketscript

[Rash: The Reckless Racket Shell](https://docs.racket-lang.org/rash/) - unique example of a shell language 

[Riposte](https://docs.racket-lang.org/riposte/) - a working DSL relevant to modern web programmers. also an implantation of an external DSL IIRC.

[Scratchy: A Scratch-like Toy](https://docs.racket-lang.org/scratchy/) - learning example 

[Syndicate](http://syndicate-lang.org/) - I believe this started on Racket and now has implementations on rust, java, type/javascript, python, squeak and Racket. 

[Video: A Language for Making Movies](https://lang.video/) - I love this example but it is currently non-functional - I believe it is considerable work to update links to ffmpeg and other tools - I really want to keep it but being non functional is problematic.

Qi - the only example of a flow language and along with Rhombus is one of the most active 'Lang communities' within the racket family.

Forth - historically important and only example of a concatenative language

ActorBasic - is this too close to Syndicate? I'm inclined to think Syndicate is something different.

Rhombus - has to be included

Typed Racket - needs an entry - sadly only gets a passing mention on the homepage.

minikanren - another book lang - I think these should be a given.


https://racket-lang.org/languages.html

mfelleisen:
@spdegabrielle  You write "I think this should be a page to showcase interesting languages with as much variety as possible" so you and I are on the same page here. You added *much* to my sentence and that's fine. In my experience though people tend to confused *quantity* with *variety* -- two rather different concepts. Let's not do that. 

Your list also hits on a rather important point with "being non functional is problematic". I agree that languages listed on the Racket _home page_ should run out of the box. How does it look when first-time visitors want to explore a listed sample language and it doesn't work? 

In this spirit, I think we should keep the list _reasonably short_ and _illustrative_ of variety (syntax, computational model, semantics, targets, embedded/hosted, plus support via publications such as the miniKanren book). Why don't you make a Google spread sheet with these attributes as columns for all candidate languages for this page and then we judge? 

Of course, I can also imagine that someone (else!) makes a linked-to web page to which any community members can contribute a language. I'd favor listing such a link at the bottom of this page w/ a sentence like "For additional example, peruse the list of languages on the community contribution page."

spdegabrielle:
Thanks @mfelleisen 

> languages listed on the Racket home page should run out of the box. 

In addition to video, it looks like a few will need to be dropped at least for now; 
* Lindenmayer fails to build (due to pict3d failing afaict)
* Hackett fails

> In this spirit, I think we should keep the list _reasonably short_ and _illustrative_ of variety (syntax, computational model, semantics, targets, embedded/hosted, plus support via publications such as the miniKanren book). Why don't you make a Google spread sheet with these attributes as columns for all candidate languages for this page and then we judge?

It's a good idea but I'm under qualified.

I've decided to outsource this as much as possible. 
https://racket.discourse.group/t/racket-languages-showcase-nominate-a-language/2536?u=spdegabrielle

s.

rfindler:
> In addition to video, it looks like a few will need to be dropped at least for now;
> 
>     * Lindenmayer fails to build (due to pict3d failing afaict)

Lindenmayer builds successfully for me. On the package server, lindenmayer just says "killed" after about 45 seconds so I'm not sure why; maybe it ran out of memory?. pict3d also seems to be killed but the pfds dependency has a test failure. Anyway, I don't think lindenmayer fails to build in general.

spdegabrielle:
currently plan - create a documentation only package that showcases current languages in the Racket ecosystem

stretch goal - do it programatically so languages are added automatically if put on the package catalog.

spdegabrielle:
and change this from a pr to a issue.

PR #294, opened 2023-12-06
ci: add a workflow to push to racket-lang.org

sorawee:
This will require adding a secret to the GitHub repo setting under the key `AWS_KEY`.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -1,0 +1,31 @@
+name: Push to racket-lang.org
+
+on: workflow_dispatch
+
+jobs:
+  coverage:
+    name: Build and push the site
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Racket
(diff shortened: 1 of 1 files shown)


sorawee:
Right now the workflow is set to run on `workflow_dispatch`. That means it requires an explicit click in the "Actions" page to deploy. 

Screenshot 2023-12-06 at 11 58 12 AM

Alternative approaches include: run on push to master branch (probably a terrible idea), run on push to a release or deploy branch.

rfindler:
Forgive my ignorance: how does this keep our secrets secret?

sorawee:
There's a form on GitHub repo settings that allow us to specify the secret. It is then saved to GitHub's system, but the value is never shown to the UI again.

Then, in a workflow, we can query for this secret value and use however we want. GitHub will automatically redact the secret from the log if there's an attempt to log the secret though.

Obviously, it's still possible to leak the secret by e.g., sending it to an external server, or mangle with the secret so that GitHub can't recognize it and thus can't redact it. But if we don't do anything crazy, these scenarios shouldn't happen.

PR #295, opened 2023-12-20
switch Mac download default to Apple Silicon

mflatt:
Continuing from #291. The initial commit here just swaps the default from Intel to Apple Silicon, and it shows an alert for "Intel Mac users" when Apple Silicon is selected (side-stepping the question of whether to mention M1, etc.).

See #291 for JavaScript strategies to try to pick the variant that matches a user's machine. I haven't tried to incorporate that here, because it seems too complex and fragile to me, but I don't object if someone else wants to add to this PR.

### DIFF
--- a/download/download-pages.rkt
+++ b/download/download-pages.rkt
@@ -556,8 +556,8 @@ var property = null;
       else if (l("WOW64")) return [Win64, Win];
       else if (l("Win"))   return [Win32, Win];
       else if (l("Mac"))   return [
-        l("Intel") ? MacIntel64 : MacPPC,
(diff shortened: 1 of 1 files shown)


sorawee:
This looks good to me. I'd love to have this merged and deployed before 8.12 release.

sorawee:
Is this OK for merge?

mflatt:
Still seems like a good idea to me. (I don't think it has to be on a release boundary, and it can work at any time.)

sorawee:
Well, release is when people will use this page a lot, so I think it's good to get this in soon.

PR #296, opened 2023-12-29
realm book fix broken link

spdegabrielle:
https://www.realmofracket.com/ offline
change to publisher page https://nostarch.com/realmofracket.htm


sorawee:
https://www.realmofracket.com/ is now back. Should we revert this PR?

mfelleisen:
Let’s use two links: one to the site and keep the back up that Stephen put in.

PR #303, opened 2024-05-13
Update individuals.html.pm

spdegabrielle:
typo and add "Sage Gerard"


spdegabrielle:
This does not acknowledge all new contributors since the page was last updated in 2018. it only adds one contributor we do know about.

PR #307, opened 2024-06-26
rcon 2024: Get started

jessealama:
Here's an utterly minimal version of the RacketCon 2024 page. This is very much something that we can iterate on. I think it makes sense to try to get something out there, even a skeletal page, so that we can start linking to it in other places, such as in Discord and on Discourse, etc. Some essential information is still missing, such as a registration page and suggestions for hotels. I'm also not quite sure if the venue is correct.

### DIFF
--- a/rcon/2024/.gitignore
+++ b/rcon/2024/.gitignore
@@ -1,0 +1,1 @@
+/index.html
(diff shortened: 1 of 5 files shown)


mflatt:
I updated some info and adjusted the style to make it more distinct from other years (which I find helpful to make sure I'm looking at the right year).

PR #309, opened 2024-07-05
rcon: Add Jack Firth's talk

jessealama:
Also, insert a coffee break after the keynote (assumed to be one hour long).


jessealama:
cc @jackfirth

PR #382, opened 2026-05-24
Update team.html.pm to remove SFC reference

spdegabrielle:


mflatt:
This looks good to me - thanks!

I didn't push the button to update the web site, because I worry that v9.2 changes may be in progress. I expect it will get updated on the live page soon, though.

PR #384, opened 2026-05-25
RC16 Website preparation

aanastasiou:
Preparing the website for RC16. New template, updated content.


mflatt:
This looks like a fine setup, so no real objection from me.

I think it would be good to adjust the style to make it different than last year — even just changing the color scheme can help distinguish that year (at least, I tend to get lost otherwise), but consider changing fonts and layouts, too.

aanastasiou:
@mflatt thank you, I appreciate the flexibility. 

My perception is that the current layout of the website is great because it is simple and hierarchical enough to direct the viewer's attention to the important points. But I also agree with the need to differentiate between years. 

For this revision, I tried to incorporate the colours of [the organising state's flag](https://en.wikipedia.org/wiki/Oakland,_California#/media/File:Flag_of_Oakland,_California.svg). Along with Racket's logo colours, this makes 4 almost matched colours. I also removed the pure blue and pure white choices that were very harsh (to my eyes at least) and went for a  [Lexend](https://www.lexend.com/), since the content is text heavy.

I am not sure how you guys would feel about applying a subtle change to the logo, along the lines of "annual modifications". Again, I am leaning on the organising state's flag here as it is a tree. Here are two brief proposals which can be worked on a little bit more if we wanted to go down that route:

image

image

Having seen the code that generates the conference pages, I would like to suggest treating these few scripts as static generators for the needs of the conference (Call for papers, schedule, "archival" mode) since we are going to be needing them every year. For example, the schedule content could be uploaded as a CSV and generate the layout via code. And while doing this, I could also see how to incorporate a basic template mechanism so that each year's design can be changed more easily via external files.

What do you think?

mfelleisen:
@aanastasiou  What do you mean with " the organising state's flag"?  

(We're not a state, and we don't have a state flag :-)

rfindler:
@mfelleisen This link goes to ~~the state of california's~~ the city of Oakland's flag. 😄

mfelleisen:
@rfindler  Yes, I followed the link before I posted .. but why is it "the organization's flag"?

aanastasiou:
> @aanastasiou What do you mean with " the organising state's flag"?
> 
> (We're not a state, and we don't have a state flag :-)

I understand. My mistake. What I meant was something that would be reminiscent of the place where the conference is organised.

spdegabrielle:
@aanastasiou thank you so much for this, and my apologies for not replying sooner.

I love the inclusion of the Oakland oak motif: It brings to mind both the Larval Stage and Texas variants at https://users.cs.northwestern.edu/~robby/logos/

I like both the 'trees on top' and the fractal(or perhaps L-system?) versions. Please go ahead. I look forward to the results.

With respect to the code generation approach this sounds like a great idea and you have my support and thanks for your efforts with this.

mflatt:
I agree — logo variants can be good.

I like the style tweaks in the latest candidate. Should we go ahead and merge this much, so that the web page will have the 2026 date and place?

aanastasiou:
@mflatt I added the 2026 logo to the directory and the page.

I am also working on the changes I suggested earlier, but certain macros are giving me a bit of a hard-time right now.

In any case, if there is agreement on the design of the website I would suggest we go forward with it and even when I push the changes, it would still preserve the agreed design for 2026. 

Could I please ask if the logo is legible for others as well? I have tried a few different combinations but if it does not look right in someone's configuration please let me know to adjust it.

mflatt:
This current version is now live.

I took the liberty of refining the location information at the top of the page, because "Oakstop, Oakland \n Oakland, ..." seemed repetitive, but feel free to revert that (commit 9030e7d1f2e69d9a75d026a84e2aacf240505258) if it was a bad idea.

spdegabrielle:
Thank you @aanastasiou - the new site looks awesome 👏👏👏

aanastasiou:
@spdegabrielle Ah, you are too kind but thank you anyway :) . 
I am working on the follow up to generate the schedule but let me know for any additional information as it becomes available and I will be adding it to these pages.

PR #390, opened 2026-06-24
Fix typo in Rhombus 1.0 announcement

gus-massa:
I'm not sure if typos in the blog post are fixed, but in case they are fixed, I remove a duplicated "the".


mflatt:
Thanks!

PR #397, opened 2026-08-23
more RacketCon'26 talks

mflatt:
@aanastasiou 

@spdegabrielle Are the name and title right for JJ's entry?


aanastasiou:
@mflatt I was able to home-in on their [personal website](https://toki.la/) but they keep referring to themselves as JJ throughout. The website includes a link to codeberg and eventually cio. I am intrigued...but not very useful. Had a brief read through both abstracts, did not spot any issues.

PR #398, opened 2026-08-29
Minor layout change

aanastasiou:
Added another class 'distinct' to give enough white space around certain messages such as the registration link and merged Call for Presentations with Registration.

I felt that we should still leave a hint for people to send in talks.


mflatt:
How about "this" on the first line and then "registration link" as the big text?

PR #400, opened 2026-09-03
add Sam Phillips talk

mflatt:
Also, add affiliation in bio for Matthew.

@aanastasiou


aanastasiou:
Thanks, cannot spot anything amiss

PR #401, opened 2026-09-03
RC16 background revision

aanastasiou:
Incorporated RF, MF image to the design of the website (recolor pattern, compose image). Clean λ, radial blur for focus. (All material in resources/), `resources/bground.png` is the final image but `bground_opt.png` is the optimised version.

What do you guys think?


mflatt:
I like this a lot!

mflatt:
I will post another PR with one more talk (at least the speaker and title). With that, we'll have a full program, and it would be good to advertise one more time about registration.

Early registration ends September 13, so maybe it would be good to highlight that on the web page, remove the call for talks proposals, and post this background change all at the same time?

@spdegabrielle

rfindler:
This looks amazing!

PR #404, opened 2026-09-13
Schedule refinement

aanastasiou:
Added opening times, clarified Saturday's lunch arrangement.

Also, made the 'special talks' headings bold (Break, Doors opening, Racket Management, Lunch). 

On my screen and eyes, their message looked a bit weak or easily skipped but have not confirmed if it looks bad on different resolutions. Feel free to revert (element definitions in `schedule.rkt`, they are based on `lecture`).


mflatt:
This looks good to me - thanks!

